PR #28, opened 2015-10-30
Component sketches

### LINKED ISSUE #11: Application scaffold - routing

### LINKED ISSUE #18: Application scaffold - actions + reducers

### LINKED ISSUE #20: Webpack + PostCSS integration
PostCSS gives us a lot of flexibility in CSS compilation. Allows for SCSS, CSS4, or whatever we want.

maxbates:
Lots of infrastructure around React + Redux + Application scaffolding, some outstanding problems. Includes:

- Application Routing (`routes.js`)
- Application Infrastructure for Actions + Reducers + Stores
  - actionCreators in `/actions/`
  - ActionTypes enumeration in `/constants/`
  - initial reducers in `/reducers/`
- Basic Application views for Dashboard, Project, Construct 
  - differentiate between smart components (`/containers/`) and dumb components (`/components/`)
  - initial components for Constructs, Blocks, Parts
- Styles / CSS
  - PostCSS integration (see webpack config)
  - decorator for components to include their own CSS (`decorators/withStyles.js`)
  - normalizer.css inclusion
  - heavy use of flexbox

close #11 
close #18 
close #20 

There are some standing errors:

- ProjectPage will error if project is not in the store. ProjectSelector (top-right corner) will add, but if you request the page directly, it will error. There should be a routing check which checks the Redux store.
- no reducers for constructs (blocks). Right now when you trigger action projectAddConstruct, it just adds constructs to the project directly. It should not. The store should be flat, such that anything with a schema is under its name, keyed by ID:
```
{
    parts : {
      <id1> : { ... },
      <id2> : { ... },
      ...
    },
    blocks : {
      <id3> : { ... },
      <id4> : { ... },
      ...
    },
    projects : {
      <id5> : { ... }
    }
}
```
- PostCSS not sourcemapped. CSS is component-only, but this is annoying.


duncanmeech:
:+1:

maxbates:
re styles: I think that generally delegating to CSS makes sense. I dont think propagating styles as component properties makes sense. Capturing some amount of state might (like a click state), but thats an arbitrary line, and classes can achieve this too. I've added normalize.css, but adding all of bootstrap is a lot of overhead. Can we wait until the design firms up a little, since we're just wireframing right now?

+1 for JSdoc documentation, could be better about this

+1 for .editorconfig file or something. I prefer 2-space indents. Sorry, these all look the same for me at 2 spaces, but I think sublime and webstorm handle indents differently.

Otherwise, good to go? feel free to merge now, would like to start this week with all these merged in if ok so we dont have too crazy a tree

maxbates:
I haven't tried linting at all, but I will go through and clean things up

maxbates:
ooh - one thing. Actions + reducers not fully updated yet for this tree structure. So the page will render, but things like "add new construct" will be buggy. if people are devving, maybe just merge in and will fix shortly, since would rather them not have to merge all this in.

maxbates:
Ok. actions + reducers are updated. 

someone shoudl make issues for:

- linting
- JSDoc format
- .editorconfig or something similar. usually they can pull in jsHint or JsLint too.

PR #43, opened 2015-11-03
Bootstrap css

duncanmeech:
Add bootstrap via sources using the bootstrap-webpack module.

bootstrap-webpack is a nice package that provides all the webpack loaders for integrating bootstrap.
Additionally it provides a .bootstrap.config.css and .bootstrap.config.less in the project root that you can use to overwrite the default styles ( I've added a custom theme just to show how this works ).

Additionally this PR styles the global nav and project selector using bootstrap and adds a bootstrap test page at /bootstrap.


maxbates:
can we move this scss, and we need to port our existing css to be compatible

maxbates:
old

PR #44, opened 2015-11-03
Adding nightwatch as the e2e framework along with a single sample test

duncanmeech:
Adding nightwatch for e2e tests.

- Install Java 8 JDK ( it is possible to do this with npm but it can get tricky )
- npm install
-  ./node_modules/selenium-standalone/bin/selenium-standalone install

Then you should be able to run 'npm run nightwatch' to run all the tests.


maxbates:
cool +1

Issue #21, opened 2015-10-29
Review + Enforcement of Styleguides


maxbates:
This is what we're using now:

[ES6](https://github.com/airbnb/javascript)
[React](https://github.com/airbnb/javascript/tree/master/react) 
[CSS/SCSS](https://github.com/airbnb/css) if we want that 

Should add a Styleguide.md to project root, listing these, and other conventions for the app

The airbnb packages include linting files

PR #45, opened 2015-11-04
fix npm run lint command, fix lint errors and add comma-dangle rule

duncanmeech:
- fix the npm run lint command ( needs to ensure a zero exit code to avoid npm error ).
- fix lint issues for three rules in .eslintrc

To check your commits:
npm run lint


maxbates:
see #21

maxbates:
see #53

PR #54, opened 2015-11-10
Model classes

maxbates:
Very simple port to using models instead of arbitrary generators. From here:

- models could become immutables, and all transformations will just return a new object
- models could trigger actions themselves (reliant on our event architecture)
- other?


maxbates:
Think we're going the route of these models being immutables. All mutations will return a new object. The actions / reducers will call functions like `addComponent` and get the new version of the object, which is kept in the store. @dchandran any opinions?

duncanmeech:
:+1:

PR #59, opened 2015-11-12
add vector2d and tests

duncanmeech:


maxbates:
Can we strip out underscore? It looks like very little of it is being used.

PR #60, opened 2015-11-13
Dontmutables

### LINKED ISSUE #23: Immutable.js - investigate / incorporate
Likely want something like this early on to benefit from memory-allocation shared data structures. Will affect how all reducers work, so should do this early

maxbates:
avoiding the use of immutable library and version tags. Instances can be accessed and mutated directly, but you should use the `mutate()` method (or create your own) which return new instances. This approach is redux compliant, should be fairly speedy, and is the most developer-friendly when used properly.

Not using immutable.js because classes cannot be extended, so there is not a clear way to create instance methods except to use static methods for the class, and conversion to-from immutables between components-store adds overhead. We can reassess if this is slow.

close #23

Not using version tags with mutable instances largely because redux doesn't place nice with mutables, but also memoizing is not straightforward for react components, and basic reassignment should not add much overhead.


duncanmeech:
Is it possible to add Object.freeze to prevent accidental changes to our immutables?

maxbates:
see #63 for freezing - will do this sprint

PR #91, opened 2015-11-25
Server API and Extensions API

dchandran:
To test the REST API from "outside", run
> npm run start 
> python server/api/api_test.py 

The above test includes both the basic REST API and Extensions API

To test the functions used within the REST API, run
> npm run test-server

The above test includes both the basic REST API but not the Extensions API yet

The REST API is documented here:
https://github.com/autodesk-bionano/genome-designer/wiki/REST-API

The Extensions API is documented here:
https://github.com/autodesk-bionano/genome-designer/wiki/Extensions-API


maxbates:
other things:
`npm run test` now runs server tests, you dont need the separate command

Is there anything relevant to extensions in the code?

package.json
```diff
@@ -78,13 +84,13 @@
     "react-transform-catch-errors": "^1.0.0",
     "react-transform-hmr": "^1.0.0",
     "redbox-react": "^1.0.1",
+    "redis": "^2.3.0",
     "redux-devtools": "^2.1.5",
     "rimraf": "^2.3.4",
     "selenium-standalone": "^4.7.1",
-    "sinon": "^1.17.2",
```
maxbates: why were `sinon` and `sinon-chai` removed? They are still needed....

server/authentication.js
```diff
@@ -0,0 +1,36 @@
+import { get as dbGet, set as dbSet } from './database';
+const crypto = require("crypto");
+
+/**
+ * @description asserts valid session key
+ * @param {sha1} session key
+ */
+export const validateSessionKey = (key) => {
```
maxbates: You probably want the syntax like:
```javascript
return new Promise((resolve, reject) => {
  dbGet(key).then(resolve, reject);
});
```
with your error handling as appropriate

server/authentication.js
```diff
@@ -0,0 +1,36 @@
...
+        password : pw, 
+        permissions: 4,
+        expiration: -1
+    };
+
+    return dbSet(sha1, info).then(result => {
+        return Promise.resolve(sha1);
+    }).catch(err => {
+        console.log("Session Key could not be saved!")
+        return Promise.reject("Session key could not be generated");
```
maxbates: promises should reject with errors. Please add your errors to the error constants file to keep consistent.

server/getRecursively.js
```diff
@@ -0,0 +1,79 @@
...
+
+        return getRecursively(next, field, (recursionDepth - 1), result);
+      }));
+    })
+    .then(() => result);
+}
+
+export const getParents = (instanceID) => {
+  return getSafe(instanceID).then( instance => {
+    return getRecursively(instance.components, 'parent');
```
maxbates: this is wrong

server/history.js
```diff
@@ -0,0 +1,103 @@
...
+  const ancestryKey = makeHistoryKey(instanceId);
+  return getSafe(ancestryKey, {})
+    .then(history => {
+      return history.ancestors || null;
+    });
+};
+
+export const getImmediateDescendants = (instanceId) => {
+  const descendencyKey = makeHistoryKey(instanceId);
+  return getSafe(descendencyKey, {})
```
maxbates: the default value of line 31 should just be saved as a constant and used as the default value here as well, avoid the `|| []` syntax

extensions/index.js
```diff
@@ -0,0 +1,97 @@
...
+});
+
+router.post('/run/:id', jsonParser, (req, resp) => {
+  const { id } = req.params;
+  const inputs = req.body;
+  const dir = getNodeDir(id);
+  const key = req.headers["session-key"];
+  var outputFiles = {};
+
+  validateSessionKey(key).then( valid => {
```
maxbates: This should be likely be run as middleware, or put in the router as a catch, not be run within specific endpoints

maxbates:
other things:
`npm run test` now runs server tests, you dont need the separate command. we can remove from package.json

BIG THINGS

- **please use JSDoc comments for non-test code, particularly things that other people will be looking at**
- **please run `npm run lint` and fix errors to keep code consistent**

These will add some work, but should happen before merge

PR #97, opened 2015-12-01
Server api

maxbates:


maxbates:
Your docker scripts do not work for me becuase they do not specify directories. Also, not sure how to run the final command because the shell is still hanging after the run (running server doesn't return you to the command prompt). after this is merged in, please create an npm script which runs these commands. see #103. I cannot test the python scripts with the code you gave me.

PR #102, opened 2015-12-02
server-api was not passing some tests. this pull request is the fix

dchandran:
Tests server inside Docker:
> docker build -t "genome-designer" 
> docker run -i -p 3000:3000 -t genome-designer
> python server/api/api_test.py 

Internal tests
> npm run test


maxbates:
and there are some reversions that we made in your first commit, beyond adding all the authentication. This would cause problems + is not valid as is....

for authentication, there should be middleware, not a check in every api endpoint. but again, should not be part of this pull request

maxbates:
I think we should also clarify the purpose of the python tests. THe majority of those tests should be run in javascript. There should be some run in python perhaps, but not unit-like tests.

server/getRecursively.js
```diff
@@ -63,6 +62,12 @@ function getRecursively(ids = [],
     .then(() => result);
 }
 
+export const getParents = (instanceID) => {
```
maxbates: why was this added again?

server/history.js
```diff
@@ -28,23 +28,24 @@ export const createDescendant = (instance) => {
...
+  //todo - need to handle ACID assurances
   const ancestryKey = makeHistoryKey(childId);
   const descendantsKey = makeHistoryKey(parentId);
 
   return getSafe(descendantsKey, {id: parentId, ancestors: [], descendants: []})
     .then(history => {
       const { ancestors, descendants } = history;
-
-      //Ancestors consists of the entire history of the instance, i.e. all parents up to the root
+
```
maxbates: why did you delete these comments

test/server/REST/history.spec.js
```diff
@@ -21,7 +21,7 @@ describe('REST', () => {
     describe('/ancestors', () => {
 
     });
-    describe('/descendants', () => {
+    describe('/descendents', () => {
```
maxbates: wrong

maxbates:
Your docker scripts do not work for me becuase they do not specify directories. Also, not sure how to run the final command because the shell is still hanging after the run (running server doesn't return you to the command prompt). after this is merged in, please create an npm script which runs these commands. see #103

PR #109, opened 2015-12-04
Sbol inventory

### LINKED ISSUE #107: SBOL inventory

maxbates:
close #107 
address #86 

Very basic version of the inventory. Does not have symbols yet. But data events, how it will be set up, etc. are ready - check SketchConstruct for an example.


maxbates:
Ready to go! still a little rough, but has images + functionality etc.

dchandran:
Are these svg's the ones that Anais had made?

jlachoff:
I guess there is still no way for us non-techy team members to evaluate work in progress? When I launch QA I only see the standard inventory...

maxbates:
@dchandran yes.
@jlachoff yes, using the command line, in the directory `genome-designer`
```
git fetch
git checkout origin/sbol-inventory
npm install
npm run start
```
There are a few dependencies but I think we installed them during WetLabAccelerator on your computer. Otherwise, you'll need [node and npm](https://nodejs.org/en/)

duncanmeech:
:+1:

PR #110, opened 2015-12-04
Server API for simple file IO

dchandran:
Will get replaced once the platform comes along


maxbates:
Can you write tests in javascript? Unit tests should be in javascript. an end to end test in python makes sense too.

server/api/index.js
```diff
@@ -208,6 +211,64 @@ router.post('/clone/:id', (req, res) => {
...
+  fs.readFile('./storage/' + url, (err, data) => {
+   if (err) {
+      res.status(500).send(err.message)
+   } else {
+      res.send(data);
+   }
+  });
+});
+
+router.post('/file/:url', (req, res) => {
```
maxbates: This will not pass linting...

server/api/index.js
```diff
@@ -208,6 +211,64 @@ router.post('/clone/:id', (req, res) => {
...
 });
 
+/**
+* File IO
+* Read and write files
+**/
+
+router.get('/file/:url', (req, res) => {
+  const { url } = req.params;
+  fs.readFile('./storage/' + url, (err, data) => {
```
maxbates: you probably need to add 'utf8' here unless you want to always use buffers...

maxbates:
@dchandran please run `npm run lint` before merging until we have jenkins checking these sort of things. this pull doesnt pass still

Issue #108, opened 2015-12-04
(UI) - update inventory + inspector toward spec


maxbates:
https://docs.google.com/document/d/1S55Ddn5rAbBdw-6N_6nvomB4efmvy8B6OeO5nruxFNE/edit#

PR #113, opened 2015-12-05
initial inspector + update inventory styling

### LINKED ISSUE #99: (inspector) Initial version of inspector + actions/reducers

maxbates:
add initial version of inspector+actions+reducers. update inventory+inspector styles. update projectPage chrome with inspector.

address #108
close #99


maxbates:
@duncanmeech check the new projectPage for some changes to incorporate the inspector, and SketchPart for how to send the current instance to the inspector. RIght now it just takes the instance itself. Might make sense to convert it to getting it from the store instead, to handle changes...

PR #116, opened 2015-12-08
Minor change - moved history test case from api_test.py to history.spec.js

dchandran:
npm run test - all pass, including new test case


maxbates:
I would suggest using the scaffolding function for these tests, not the hardcoded one. We have a static hardcoded one in /test/schemas/_examples

PR #121, opened 2015-12-09
Genbank parser as a "Node" extension

dchandran:
This pull request has the extensions framework and has the genbank parser as an example. The extensions framework uses the same interface as the platform that Dion is developing. The genbank parser is implemented in Docker so that the main server does not need to install python and biopython (and everything else that each extension would require) - it keeps the requirements modular.

The /extensions/compute folder has all the "node" extensions.
The /extensions/foundry folder has the "toy" foundry API example and all of E.coli genes as an example inventory (block data format).

@IsaacLuo check if the code below is sufficient for Onion.

Below is the code that Onion would need to execute. The first REST call is for executing the genbank parse code. The second is to update the block in the server database because the parser program just returns the block object. The code is basically two REST calls.
```
//save selected block ID
var block_id = block.id; 

//tell the extension what to use as input and
//where to write the output
//any output that does not have a file specified 
//will be returned in the json

var inputs = {
  "genbank":"extensions/compute/genbank_to_block/sequence.gb", 
  "sequence":"storage/block/"+block_id+"/sequence"
};

//Call the genbank extension
var output;
var onion_data;

//don't know why I always need double 'then' 
gd.runProcess("genbank_to_block",inputs).then(
    function(resp) { return resp.json(); }).then(
        function(resp) { output = resp;
            block = JSON.parse(output.block);    
            data = JSON.parse(output.onion_data); 
         }));

//We need to update block sequence URL
//because the genbank extension currently
//does not write anything to the database

//can replace the call below with saveBlock function
block.id = block_id
block.sequence.url = inputs.sequence

fetch(api_url + "block/" + block_id, {
        method: 'put',
        headers: headers,
        body: JSON.stringify(block)
}).then( function(resp) {
    return resp.json();
}).then( function(resp) {
    console.log("success");
});
```

The above code assumes that you already have the session-key and a block. Here is an example code for generating those. After doing npm run start, you can run the code below in the Chrome console directly (within the app, i.e. 0.0.0.0:3000).

```
var api_url = "api/";
var run_url = "exec/";
var headers = {  "Content-type": "application/json" };
var block;
var input;
var output;

/*****
You need to do this once, just to get a session key
The client side would cache this and automatically
append session-key to all requests 
******/

fetch(api_url + "login?user=''&password=''", {
        method: 'get',
        headers: headers
}).then( function(resp) {
    return resp.json();
}).then( function(resp) {
    headers["session-key"] = resp["session-key"];
});

/*********
Create a block. In the app itself, Onion would not need
to create a block because a block will already exist, 
so this step is not needed.
*********/

block = {
  "metadata": {
    "authors": [],
    "version": "0.0.0",
    "tags": {}  },
  "options":[],
  "components":[],
  "rules": [],
  "notes": {}
}

fetch(api_url + "block", {
        method: 'post',
        headers: headers,
        body: JSON.stringify(block)
}).then( function(resp) {
    return resp.json();
}).then( function(resp) {
    block = resp;
});

/*********
This is what Onion would do
**********/

#save block id
var block_id = block.id; 

#tell the extension what to use as input and
#where to write the output
#any output that does not have a file specified 
#will be returned in the json

var inputs = {
  "genbank":"extensions/compute/genbank_to_block/sequence.gb", 
  "sequence":"storage/block/"+block_id+"/sequence"
}

#Call the genbank extension
var output;
var onion_data;

fetch(run_url + "genbank_to_block", {
        method: 'post',
        headers: headers,
        body: JSON.stringify(inputs) #input
}).then( function(resp) {
    return resp.json();
}).then( function(resp) {
    
    #get the block and onion_data from the results
    output = resp;
    block = JSON.parse(output.block)    
    data = JSON.parse(output.onion_data)

});


#We need to update block sequence URL
#because the genbank extension currently
#does not write anything to the database

block.id = block_id
block.sequence.url = inputs.sequence

fetch(api_url + "block/" + block_id, {
        method: 'put',
        headers: headers,
        body: JSON.stringify(block)
}).then( function(resp) {
    return resp.json();
}).then( function(resp) {
    console.log("success");
});
```

Python test case that tests the above scenario (need to move these to JS at some point).
```
$ipython server/api/api_test.py 
----------------------------------------------------------------------
Ran 7 tests in 11.664s

OK
```


Dockerfile
```diff
@@ -9,14 +9,28 @@ RUN wget http://download.redis.io/redis-stable.tar.gz
...
+RUN apt-get install -y --force-yes docker-engine
+
+#setup node
 ADD package.json /app/package.json
 RUN cd /app && npm install
 RUN npm update -g npm
 
 EXPOSE 3000
-ENV PORT=3000
+#ENV PORT=3000
```
maxbates: You probably want to keep this to match line above

maxbates:
You should add these new directories to be linted... add them to the lint function in package.json

extensions/compute/genbank_to_block/read_genbank.py
```diff
@@ -0,0 +1,90 @@
...
+      "metadata": {
+        "authors": [],
+        "version": "0.0.0",
+        "tags": {}  },
+      "options":[],
+      "components":[],
+      "rules": [],
+      "notes": {},
+      "sequence": { 
+        "url": None,
```
maxbates: are we using the field ID or url? im fine either way.. need to update /schemas/Sequence though.

maxbates:
There will be functions for interacting with the block. You cannot do direct assignments like your examples. I am going to freeze blocks soon, so that you have to mutate them using the API, to maintain guarantees of immutability.

maxbates:
Setting sequence should also set the length field on the sequence

maxbates:
This API is exposing too much about weird things on the server and is very brittle. Why the absolute paths to issue the request? 

in my onion-prep branch, i have renamed the header `session-key` to `sessionkey`

please move all tests to JS asap. I am hesitant to merge this without some tests in JS that are easy / quick to run + share code base / utils with the other API functions

maxbates:
there should be some helpers added to middleware/api to run extensions from the client

src/index.js
```diff
@@ -5,7 +5,7 @@ import { ReduxRouter } from 'redux-router';
 import * as actionTypes from './constants/ActionTypes';
 import actions from './actions/_expose';
 import store, { lastAction } from './store/index';
-import { login } from './middleware/api';
+import { login, runProcess } from './middleware/api';
```
maxbates: wrong

src/middleware/api.js
```diff
@@ -7,7 +7,9 @@ import BlockDefinition from '../schemas/Block';
  Utils
  *************************/
 
-const serverRoot = 'http://localhost:3000'; //fetch only supports absolute paths
+const serverRoot = ''; //fetch only supports absolute paths
```
maxbates: why? this will make tests fail

src/schemas/Block.js
```diff
@@ -19,13 +19,26 @@ import SequenceDefinition from './Sequence';
  In addition to sequence annotations, a block may list `notes`, which are essentially annotations that do not specifically reference the sequence.
  */
 
+const ruleShape = validators.shape({
```
maxbates: All changes in this file are reversions and incorrect

src/middleware/api.js
```diff
@@ -78,6 +80,15 @@ export const saveBlock = (block) => {
   }
 };
 
+export const runExtension = (id, inputs) => {
```
maxbates: This should have explicit JS tests, not just testing API endpoints

extensions/compute/api.js
```diff
@@ -0,0 +1,199 @@
+import express from 'express';
+import bodyParser from 'body-parser';
+import { runNode, getNodeDir, buildNodeContainer } from './runNode';
+import { sessionMiddleware } from '../../server/authentication';
+
+const mkpath = require('mkpath');
+const fs = require('fs'), path = require('path');
```
maxbates: this file has several lint errors

extensions/compute/api.js
```diff
@@ -0,0 +1,199 @@
...
+            }));
+        });
+
+        //outputs
+        for (i=0; i < outputs.length; ++i) {
+          outputFiles[ outputs[i].id ] = "";
+          outputFileNames.push(dir + "outputs/" + outputs[i].id);
+        }
+      
+        Promise.all(inputFileWrites).then(result => {
```
maxbates: probably need to return. and should avoid nesting if possible

extensions/compute/api.js
```diff
@@ -0,0 +1,199 @@
...
+
+                    }
+                  }); //make path
+                } else { //if not writing to output file, return in json
+                  outputFiles[ outputs[i].id ] = contents;
+                }
+              }
+
+              //return object
+              resp.json(outputFiles);
```
maxbates: should wait for files to be written...

maxbates:
also, we probably want to expose all the middleware in some way - not just run extension? do we want to give people ability to write / read files? if so then we should expose those as well (e.g. if you ened to upload a genabnk file)

dchandran:
@maxbates merged master again. aside from lint errors which we can deal with later, I don't think there are any merge issues now

src/index.js
```diff
@@ -5,7 +5,7 @@ import { ReduxRouter } from 'redux-router';
 import * as actionTypes from './constants/ActionTypes';
 import actions from './actions/_expose';
 import store, { lastAction } from './store/index';
-import { login } from './middleware/api';
+import { login, runExtension } from './middleware/api';
```
maxbates: Probably need to expose this middleware better - we can expose the whole module I think?

test/models/Block.spec.js
```diff
@@ -16,11 +16,12 @@ describe('Model', () => {
     describe('Annotations', () => {
       const annotation = AnnotationDefinition.scaffold();
 
+      /*
```
maxbates: Can you explain what was erroring? I don't like just deleting a test to hide the problem... I think ID should be required on each annotation, I dont know why you had to delete it...

dchandran: I made ID optional for the annotations schema. Why is ID required? When parsing genbank, having to add a uuid to each annotations seems unnecessary. But it can be done if there is a good reason.

maxbates: I suppose its not. It makes data manipulation a lot easier when each has an ID. I dont see a good reason not to?

dchandran:
@maxbates alright, I put the required back in for Annotation schema. Just had to adjust the genbank code to add uuid to each feature.

maxbates:
to aggregate:

- [ ] expose all the middleware, not just one function (`import * as middleware from '../src/middleware/api` and expose `middleware`)
- [ ] test for runExtension() function, not hitting REST directly 
- [ ] put together an example that can run from the window
- [ ] compute/api.js should return after all files written (see other comment)
- [ ] issue to fix lint errors, noting all files
- [ ] move your python tests to the test folder, not the server folder

Also, it would be nice if the extensions / compute routers existed in the server directory directly, and simply referenced other folders as needed (specifically extensions - compute is only temporary so whatever)

dchandran:
@maxbates not sure how you want to implement "expose all the middleware, not just one function".  just add them to exposed in index.js?

maxbates:
Yes. The same way ActionTypes is exposed.

maxbates:
See also question regarding promise resolution timing for returning the URLs

dchandran:
@maxbates all the items are done except lint errors

maxbates:
issue now is 159

maxbates:
docker tests are still failing for me

PR #130, opened 2015-12-10
add sbolType action + model method. convert block.rules to object fro…

maxbates:
…m array

address #86


src/actions/blocks.js
```diff
@@ -18,11 +19,20 @@ export const blockCreate = (initialModel) => {
...
 
-export const blockClone = (blockId) => {
+//this action accepts either the block directly, or the ID
+//inventory items may not be in the store, so we need to pass the block directly
+export const blockClone = (blockInput) => {
   return (dispatch, getState) => {
-    const oldBlock = getState().blocks[blockId];
-
-    //hack - should hit the server with fetch()
+    let oldBlock;
```
maxbates: Basically, this is because these actions all expect the block to be in the store - since things in the store are block models, not just JSON blobs I dont think this is great, but its a big refactor so would need to revisit beginning of the next milestone -- not just because they have to be models, but also because the actions all expect the block to be in the store.

PR #140, opened 2015-12-11
uuid and versions in server api

dchandran:
Clone and update (PUT) in the server api will store the same data using two different keys:
First key is <uuid>:<version>
Second key is <uuid>

This doesn't break existing code because using the id by itself will return the latest block or project.
Getting a previous version can be achieved by using uuid:version, e.g.  /api/project/3407154b-31dc-4f6d-b44d-1acafd9bce1d:0.0.0 

Test case in api_test.py


maxbates:
This should have several javascript tests, not just an end to end test. 
Infrastructure for versioning should be in javascript, outside of the API router.
inconsistent use of instance.version and instance.metadata.version
lint errors. especially prefer const to let.
GET routes are a bit ambiguous. I don't think POST and PUT setting the head themselves is the safest way of achieving this... at least in the short term let's abstract that out so the logic is shared / easier to update.

PR #146, opened 2015-12-15
Deep Object.freeze on models

maxbates:
Enables deep freeze (Object.freeze) on models when not in production, to ensure there is no attempt at direct property mutation. 

Small updates to how mutations are done, updates to tests / inventory to avoid using Object.assign and use immutable-friendly API instead


duncanmeech:
:+1:

PR #152, opened 2015-12-16
Inventory andrea

maxbates:
Adds Andrea's parts to the inventory, and uses them as the default parts in the inventory. Creates a sequence file for each. inventory items now have pinned IDs (but will be cloned when dragged)


maxbates:
@IsaacLuo you can merge this branch when ready and all parts from inventory will have a sequence now (but no annotations)

PR #153, opened 2015-12-16
Move login to root, restructure /server/ directory

maxbates:
Moves `/login` route to root, outside of the /api/
moves server helpers to folder /utils/ for use inside/outside of API
moves devServer to server directory
updates login middleware and tests
adds linting to server directory
several lint errors fixed


devServer.js
```diff
@@ -1,54 +0,0 @@
...
-
-// Register Client Requests, delegate routing to client
-// ----------------------------------------------------
-
-//Static Files
-app.use('/images', express.static('src/images'));
-
-//so that any routing is delegated to the client
-app.get('*', function(req, res) {
-  res.sendFile(path.join(__dirname, 'src/index.html'));
```
dchandran: @maxbates didn't notice this before merging. shouldn't this be ../src/index.html instead of src/index.html

PR #205, opened 2016-01-20
Conversion to use docker-compose

PeterRJones:
Updated Docker file and removed git commands. Submodules are working fine within Jenkins. Issue was with dokku.

Added docker-compose.yml file (for dev) and docker-compose-prod.yml (for non-dev envs). Deployment will now use this to run app. Also separated out redis into seperate container and reference it via docker link.

only line 11 in database.js has changed:
+const client = redis.createClient('redis', 6379);


dchandran:
I think you added a file - server/database.js - by accident. Can you delete it. 
Is there anything that we need to install for docker compose to work?

PeterRJones:
updated to have database.js removed from server/database.js location. effectively reverts to content in server/utils/database.js, which is okay for now.

PeterRJones:
Yes, you will need to install the docker-compose tool. That is included in the Docker Toolbox installation for the Mac installation. It is a separate package on linux. https://docs.docker.com/compose/install/

If genome designer ends up being a single container application, we can revert back to just a Docker file. If it remains needing/using multiple containers we should use docker-compose to capture this layout and the dependencies between containers.

PR #207, opened 2016-01-20
Remove docker in extension api

dchandran:
Mostly file removals. Cleaned up runNode code. Don't look at it in detail - it's not terribly important right now. Just look at test/extensions/examples.spec.js


dchandran:
This branch has been merged into the genbank branch, so no point doing the pull request

PR #208, opened 2016-01-20
import, export, and search extensions

dchandran:
The previous merge request (remove-docker-api) needs to go in before this one because it builds on the previous - that should reduce the number of files changed.

## Features added in this branch
genbank import/export
feature file import
searching NCBI nucleotide database

All of the above use the same extensions architecture, which is why they are bundled into one branch.

### test cases
Test cases are in the test/extensions folder
```
>> npm run test
172 passing (42s)
No tests failing  
```
## Overview
There are two categories of extensions introduced in this branch

### import/export
These extensions reside in the extensions/convert folder, e.g. extensions/convert/genbank and extensions/convert/features

Each extension is contained within its own folder and must define a NodeJS module (index.js) that exports one or more of the following functions: importBlock, importProject, exportBlock, exportProject

TODO: API documentation for each of the above functions. Basically, they all work with the block and project JSONs.

### search
These extensions are contained in folders within the extensions/search folder, e.g. extensions/search/nucleotide

Each search module must define one function: search, which returns a list of blocks and takes a search string and max-returns as arguments.


maxbates:
pending changes we talked about, this looks good. Ones I remember:

- manifest for each extension
- ability to list all extensions
- middleware function to run extension
- test demonstrating how plugin would list + execute extensions

Need to better clarify searching . NCBI is a special case, but we likely want to come up with a unified interface for other sources.

extensions/compute/runNode.js
```diff
@@ -1,63 +1,212 @@
...
+          }
+        }); //fs.writeFile
+      })
+    );
+
+    inputs.forEach(inputItem => {
+      const inputKey = inputItem.id;
+      inputFileWrites.push(
+        new Promise((resolve, reject) => {
+          const inputFile = fileUrls[inputKey].replace('/api/file/', 'storage/');
```
maxbates: this should be abstracted out so easier to find / dynamic

extensions/compute/runNode.js
```diff
@@ -1,63 +1,212 @@
...
-            console.log(err);
-            return Promise.reject(err);
+    return Promise.all(inputFileWrites);
+  }
+
+  function readOutputFiles(outputs) {
+    return new Promise( (resolveRead, rejectRead) => {
+      readMultipleFiles(outputFileNames, 'utf8', (err, buffers) => {
+        if (err) {
+          console.log('Output read error: ' + err);
```
maxbates: probably want to reject promise

extensions/compute/runNode.js
```diff
@@ -1,63 +1,212 @@
...
 
-export const buildNodeContainer = (id) => {
+  const fullCmd = 'cd ' + dir + '; ' + cmd;
 
-  if (nodesStartedBuilding[id]) return nodesStartedBuilding[id];
+  return exec(fullCmd).then(result => {
+    return Promise.resolve(true);
+  }).catch(err => {
+    console.log(err.message);
+    //apparently, even warning messages trigger this section of exec, so it 'usually' ok
```
maxbates: If you look at https://github.com/yakimchuk/promised-exec/blob/master/index.js ... you only get a rejection if there is stderr and no stdout... so im not sure this is right. maybe we could force a log? THe way node handles it, warnings go to stderr... all libraries basically rely on this

extensions/compute/runNode.js
```diff
@@ -1,63 +1,212 @@
...
+                }); //make path
+              } else { //if not writing to output file, return in json
+                outputFiles[ outputs[i].id ] = contents;
+                resolve(contents);
+              }
+            })
+          );
+        }
+
+        Promise.all(outputFileWrites)
```
maxbates: return this. shouldn't be working without it....

extensions/compute/runNode.js
```diff
@@ -1,63 +1,212 @@
...
+    cmd = data.command;
+    return createInputOutputFolders(dir);
+  })
+  .then( res => {
+    return writeInputFiles(dir, inputs, outputs);
+  })
+  .then( res => {
+    if (!res) {
+      console.log('Input write error');
+      Promise.reject('Input write error');
```
maxbates: return

extensions/convert/export.js
```diff
@@ -0,0 +1,113 @@
...
+router.use(sessionMiddleware);
+
+function exportProject(id, input, callback) {
+  if (id in extensions) {
+    const mod = extensions[id];
+    if (mod.exportProject) {
+      try {
+        mod.exportProject(input.project, input.blocks)
+        .then(res => {
+          callback(res);
```
maxbates: i'd suggest a more node-like way of handling this, rather than using an error object... many different styles: 1) node style `callback(err, res)` 2) promise 3) this format

maxbates: Personally, would prefer to use promises for all these. They make sense here

extensions/convert/export.js
```diff
@@ -0,0 +1,113 @@
...
+import { load as loadExtensions } from '../requireExtensions';
+const router = express.Router(); //eslint-disable-line new-cap
+const jsonParser = bodyParser.json({
+  strict: false, //allow values other than arrays and objects
+});
+
+const normalizedPath = require('path').join(__dirname, '.');
+const retval = loadExtensions(normalizedPath);
+const extensions = retval.extensions;
+const manifests = retval.manifests;
```
maxbates: These should be functions? so they return dynamically? imports are static

extensions/convert/features/index.js
```diff
@@ -0,0 +1,60 @@
...
+        'tags': {},
+        'version': '',
+      },
+      'components': [],
+    };
+
+    const blocks = {};
+
+    function appendToProj(data) {
+      let id, json, name, seq, type, color;
```
maxbates: you can use destructuring: `[name, seq, type, color, ...rest] = data`

extensions/convert/features/index.js
```diff
@@ -0,0 +1,60 @@
+const uuid = require('node-uuid');
+const csv = require('fast-csv');
+
+module.exports = exports = {};
+
+exports.importProject = function importProject(csvstr) {
+  return new Promise((resolve, reject) => {
+    const proj = {
```
maxbates: You should use `schema.scaffold()` rather than hardcode

extensions/convert/features/index.js
```diff
@@ -0,0 +1,60 @@
...
+
+    function appendToProj(data) {
+      let id, json, name, seq, type, color;
+
+      name = data[0];
+      seq = data[1];
+      type = data[2];
+      color = data[3];
+      id = uuid.v4();
+      blocks[id] = {
```
maxbates: again, `schema.scaffold()`

extensions/convert/features/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "features",
```
maxbates: suggest a version field, so can pin

extensions/convert/genbank/genbank_import_export.py
```diff
@@ -0,0 +1,180 @@
...
+        sf.location = SeqFeature.FeatureLocation(feature["start"],feature["end"])
+        sf.qualifiers["block_id"] = feature.get("block_id","")
+        sf.qualifiers["parent_block"] = feature.get("parent_block",None)
+        seq_obj.features.append(sf)
+
+    SeqIO.write(seq_obj, open(filename, "w"), "genbank")
+
+def create_block_json(id, sequence, features):
+    return {
+        "id": id,
```
maxbates: ideally, you would call some javascript that builds this, avoid using static models

extensions/compute/runNode.js
dchandran: This file is temporary anyway. Once the platform is ready to integrate, this file will change. I'll make the simple fixes like the ones you have pointed out, but I don't want to spend too much time on cleaning up as long as the basic API works.

extensions/convert/import.js
```diff
@@ -0,0 +1,127 @@
...
+    const mod = extensions[id];
+    if (mod.importProject) {
+      try {
+        mod.importProject(data)
+        .then(res => {
+          callback(res);
+        })
+        .catch(err => {
+          callback({
+            error: err.message,
```
maxbates: again, should avoid adding a new error handling syntax. IT doesnt seem like you handle the errors anyway.

extensions/convert/import.js
```diff
@@ -0,0 +1,127 @@
...
+import { sessionMiddleware } from '../../server/utils/authentication';
+import { load as loadExtensions } from '../requireExtensions';
+const router = express.Router(); //eslint-disable-line new-cap
+const jsonParser = bodyParser.json({
+  strict: false, //allow values other than arrays and objects
+});
+const normalizedPath = require('path').join(__dirname, '.');
+const retval = loadExtensions(normalizedPath);
+const extensions = retval.extensions;
+const manifests = retval.manifests;
```
maxbates: see note above. prefer functions, so dynamic

extensions/search/search.js
```diff
@@ -0,0 +1,61 @@
...
+router.use(sessionMiddleware);
+
+function searchString(id, str, max, callback) {
+  if (id in extensions) {
+    const mod = extensions[id];
+    if (mod.search) {
+      try {
+        mod.search(str, max)
+        .then(res => {
+          callback(res);
```
maxbates: again, avoid new syntax

extensions/search/search.js
```diff
@@ -0,0 +1,61 @@
...
+import { sessionMiddleware } from '../../server/utils/authentication';
+import { load as loadExtensions } from '../requireExtensions';
+const router = express.Router(); //eslint-disable-line new-cap
+const jsonParser = bodyParser.json({
+  strict: false, //allow values other than arrays and objects
+});
+const normalizedPath = require('path').join(__dirname, '.');
+const retval = loadExtensions(normalizedPath);
+const extensions = retval.extensions;
+const manifests = retval.manifests;
```
maxbates: again, prefer dynamic

extensions/search/search.js
```diff
@@ -0,0 +1,61 @@
...
+
+router.use(sessionMiddleware);
+
+function searchString(id, str, max, callback) {
```
maxbates: since you reuse this part so much, you should probably abstract into a new function that is shared

package.json
```diff
@@ -62,7 +62,8 @@
     "redux-thunk": "^0.1.0",
```
maxbates: Should be linting extensions folder (see `lint` script at top)

src/middleware/api.js
```diff
@@ -129,3 +126,67 @@ export const writeFile = (fileName = uuid.v4(), contents) => {
...
+
+export const importProject = (id, input) => {
+  try {
+    return fetch(importPath(`project/${id}`), headersPost(JSON.stringify({ text: input })));
+  } catch (err) {
+    return Promise.reject('error stringifying input object');
+  }
+};
+
+export const searchForBlocks = (id, inputs) => {
```
maxbates: can we just call this search?

test/extensions/examples.spec.js
```diff
@@ -0,0 +1,26 @@
+import { expect } from 'chai';
+import { runNode } from '../../extensions/compute/runNode';
+const fs = require('fs');
+
+describe('Nodes Extension', () => {
+  const dnaFile = process.cwd() + '/temp.txt';
```
maxbates: add to `.gitignore` -- can you add it to a testing directory if possible?

test/extensions/genbank.spec.js
```diff
@@ -0,0 +1,46 @@
...
+      expect(output.block.components.length === 3).to.equal(true);
+      expect(output.blocks['2'].components.length === 2).to.equal(true);
+      done();
+    });
+  });
+
+  it('should be able convert Genbank features to Block', function importGB(done) {
+    this.timeout(5000);
+    fs.readFile('./test/res/sampleGenbank.gb', 'utf8', (err, sampleStr) => {
+      importBlock('genbank', sampleStr, data => {
```
maxbates: can you actually check that the values are converted properly? There is no way to know that fields migrated properly

test/middleware/api.js
```diff
@@ -123,42 +122,101 @@ describe.only('Middleware', () => {
...
+    });
+  });
+
+  it('exportBlock() should be able convert Block to Genbank', function testFunc(done) {
+    this.timeout(5000); //reading genbank can take long, esp when running along with other tests
+    fs.readFile('./test/res/sampleBlocks.json', 'utf8', (err, sampleBlocksJson) => {
+      const sampleBlocks = JSON.parse(sampleBlocksJson);
+      api.exportBlock('genbank', {block: sampleBlocks['1'], blocks: sampleBlocks})
+      .then(result => {
+        return result.json();
```
maxbates: you probably want to move this to the API middleware itself, as is done for most of the other api middleware functions where relevant

extensions/compute/runNode.js
dchandran: done, at least for this specific case

extensions/compute/runNode.js
dchandran: done

extensions/compute/runNode.js
dchandran: Trivial warning still go to the catch section, so the link you sent doesn't solve the issue. Leaving it as it is for now.

extensions/compute/runNode.js
dchandran: Fixed. Files were getting written in a timely manner, hence working before.

extensions/compute/runNode.js
dchandran: fixed

extensions/convert/export.js
dchandran: changed to promises

extensions/convert/export.js
dchandran: I'm using require to load each extension module. It is done once at the start (which is why loadExtensions function is sync not async). Once we have a dynamic way of adding extensions, we can incorporate dynamically adding them to the extensions and manifest tables.

extensions/convert/features/index.js
dchandran: done

extensions/convert/features/index.js
dchandran: I don't think this is a good idea for extensions. They should be able to define the JSON directly because its simple enough. Can we auto-fill all the required fields if they don't exist?

extensions/convert/features/index.js
dchandran: See previous note

extensions/convert/features/manifest.json
dchandran: Done

extensions/convert/genbank/genbank_import_export.py
dchandran: That would complicate things. JSON-based communication is simple enough. We should auto-fill and auto-correct the incoming JSONs

extensions/convert/import.js
dchandran: Done

extensions/convert/import.js
dchandran: I'm using require to load each extension module. It is done once at the start (which is why loadExtensions function is sync not async). Once we have a dynamic way of adding extensions, we can incorporate dynamically adding them to the extensions and manifest tables.
 Add a line note

extensions/search/search.js
dchandran: Done

extensions/search/search.js
dchandran: Done

extensions/search/search.js
dchandran: I think you meant this comment for import.js and export.js, not search.js
I removed redundant code in those files.

src/middleware/api.js
dchandran: Done. Changed to search. If we have different search functions later, we can change it again - not a big issue.

test/extensions/examples.spec.js
dchandran: Done. Added temp.* to gitignore

test/middleware/api.js
dchandran: I didn't understand this comment

package.json
dchandran: added extensions to the "lint" line of package.json

extensions/convert/features/index.js
maxbates: We could on the initial creation. But I think we should validate all updates.

extensions/compute/api.js
```diff
@@ -1,215 +1,28 @@
...
-          }); //runNode
-        })
-        .catch(err => { console.log(err); }); //writeInputFiles
-      })
-      .catch(err => { console.log(err); }); //writeInputDirs
-    }   //read yaml file successful
-  }); //fs.readFile
-
+  const key = req.headers.sessionkey;
+  const header = {'sessionkey': key, 'host': 'http://0.0.0.0:3000'};
```
maxbates: Is this valid when deployed?

extensions/compute/runNode.js
```diff
@@ -1,63 +1,207 @@
...
+
+    return runCmd(id, cmd);
+  })
+  .then( res => {
+    return readOutputFiles(outputs);
+  })
+  .then( res => {
+    return Promise.resolve(outputFiles);
+  })
+  .catch( err => {
```
maxbates: You dont need this, just pass the error along by omitting a catch

extensions/convert/export.js
```diff
@@ -0,0 +1,90 @@
...
+            resolve(res);
+          })
+          .catch(err => {
+            reject(err);
+          });
+        } catch (err) {
+          reject(err);
+        }
+      } else {
+        reject('No export option named ' + id + ' for projects');
```
maxbates: different error msg

extensions/convert/export.js
```diff
@@ -0,0 +1,90 @@
...
+    } else {
+      resp.json(res);
+    }
+  })
+  .catch(err => {
+    resp.status(500).send(err);
+  });
+}
+
+router.post('/project/:id', jsonParser, (req, resp) => {
```
maxbates: Did you test these? I have had weird experiences passing the `resp` to another function, as in it didnt work

extensions/search/nucleotide/search.py
```diff
@@ -0,0 +1,40 @@
...
+sys.path.append("extensions/convert/genbank")
+
+from genbank_import_export import *
+
+db_name =  sys.argv[1]
+search_string = sys.argv[2]
+max_items = int(sys.argv[3])
+outfile =  sys.argv[4]
+
+Entrez.email = "deepak.chandran@autodesk.com"
```
maxbates: are you sure you want to commit this?

src/middleware/api.js
```diff
@@ -129,3 +126,67 @@ export const writeFile = (fileName = uuid.v4(), contents) => {
 
   return fetch(filePath, headersPost(contents));
 };
+
+/**************************
```
maxbates: Let's put these in a separate file, and import / re-export them, especially since these will later be dynamic

maxbates:
We should convert many of the middleware functions to automatically convert to JSON. for files, we can't necessarily... but for saving blocks, running conversions, etc., we should handle the `resp.json()` bit automatically (unless they will be huge, as in the case of files, so they can stream, but I dnot think we expect that)

maxbates:
If you feel you've addressed above comments feel free to merge

PR #212, opened 2016-01-21
Construct name

duncanmeech:
- Add selected construct to store
- Highlight selected construct name in white
- Fix all mouse coordinate calculations using getBoundingClientRect versus old school DOM tree walking nonsense.
- Allow SBOL blocks from inventory to be dropped between OR on blocks
- Fix some insertion marker colors.


src/containers/Inspector.js
```diff
@@ -53,8 +53,8 @@ export class Inspector extends Component {
 
 function mapStateToProps(state, props) {
   const { isVisible } = state.inspector;
-  const { currentBlocks } = state.ui;
-  const block = currentBlocks && currentBlocks.length ? state.blocks[currentBlocks[0]] : null;
+  const { currentBlocks, currentConstructId } = state.ui;
+  const block = currentBlocks && currentBlocks.length ? state.blocks[currentBlocks[0]] : state.blocks[currentConstructId];
```
maxbates: this should default to null if neither is selected

src/reducers/ui.js
```diff
@@ -3,11 +3,34 @@ import * as ActionTypes from '../constants/ActionTypes';
...
+  case ActionTypes.UI_SET_CURRENT_CONSTRUCT: {
+    // action.blocks are all the blocks in the project. We use this to de-select
+    // any selected block which is not part of the current construct
+    const construct = action.blocks[action.constructId];
+    // filter selections to including only children of the construct
+    let selectedBlocks = [];
+    if (state.currentBlocks && state.currentBlocks.length) {
+      selectedBlocks = state.currentBlocks.filter(id => {
+        if (construct.components) {
+          return construct.components.includes(id);
```
maxbates: this will need to work recursively...

PR #213, opened 2016-01-23
Lib update

maxbates:
Updates to babel 6, redux devtools

adds semicolons to several class properties (new valid syntax)

Devtools now accessible in dev by `ctrl-h`

~Bootstrap v4 forthcoming.~ Ignoring bootstrap


duncanmeech:
I'd merge this now, but @IsaacLuo needs to fix master. Then you will probably want to remerge with master

PR #214, opened 2016-01-25
Prefix ids

maxbates:
adds prefixing to IDs for schemas, to blocks, projects, a few others

dependent on #213


src/extensions/onion2
```diff
@@ -1 +1 @@
-Subproject commit 3801f08fd6493fe91d4e989935fc4f933b1c9ff6
+Subproject commit a2d950f5ea313df8af99c56be54453c3360e2150
```
dchandran: The latest commit is f8b0efaf1b1ed3aa92dcfa83b0ab381719b2e5c6 at
https://github.com/IsaacLuo/onion2/commits/master

@IsaacLuo which commit should we point to?

dchandran:
Good to merge after you remove the submodule commit change.

PR #216, opened 2016-01-26
Remove pip in package.json

dchandran:
@maxbates or @duncanmeech 

* Removed pip command in package.json. 
* Biopython needs to be installed separately for search or genbank import/export features to work. 
* Added pyc and storage to ignore list


maxbates:
Should probably go in the Dockerfile?

dchandran:
its already in the Dockerfile. I'll change it to python instead of python3 - done

maxbates:
Biopython too?

dchandran:
Yes, here are lines 23-26 of Dockerfile:
```
#everything needed by extensions
RUN apt-get install -y python python-pip
RUN yes | pip install biopython
```

PR #217, opened 2016-01-26
Menus redux

duncanmeech:
- Small clean up on main menus.
- Add popup context menus
- Add hover state to blocks in construct and launch a popup menu when clicked.


duncanmeech:
NOTE: Can't be merged until master is fixed.

PR #223, opened 2016-01-28
Modal window

duncanmeech:
Modal window widget with example attached to block construct menu.
Will need rebasing on broken master, I'll do that tomorrow but please review before that.


src/components/modal/modalbuttonbar.js
```diff
@@ -0,0 +1,21 @@
+import React, { Component, PropTypes } from 'react';
+import '../../../src/styles/Modal.css';
```
maxbates: only need to import once, recommend at container level

src/components/modal/modalwindow.js
```diff
@@ -0,0 +1,82 @@
...
+   */
+  render() {
+    // only render contents if open
+    const contents = this.props.open
+    ? (<div className="modal-window">
+        <div className="modal-window-title">{this.props.title}</div>
+        {this.props.payload}
+        {this.renderButtons()}
+      </div>)
+    : null;
```
maxbates: why not just return null always?

src/components/modal/modalwindow.js
```diff
@@ -0,0 +1,82 @@
...
+ */
+export default class ModalWindow extends Component {
+
+  static propTypes = {
+    open: PropTypes.bool.isRequired,
+    closeModal: PropTypes.func.isRequired,
+    closeOnClickOff: PropTypes.bool,
+    buttons: PropTypes.array,
+    title: PropTypes.string.isRequired,
+    payload: PropTypes.object.isRequired,
```
maxbates: can we name this more obviously

src/containers/graphics/views/constructviewer.js
```diff
@@ -200,12 +271,33 @@ export class ConstructViewer extends Component {
...
-        <ConstructViewerMenu constructId={this.props.constructId} layoutAlgorithm={this.props.layoutAlgorithm}/>
+        {menu}
         <div className="sceneGraphContainer">
           <div className="sceneGraph"/>
         </div>
+        {this.blockContextMenu()}
+        <ModalWindow
+          open={this.state.modalOpen}
+          title="Construct Viewer Modal"
+          payload={<div className="payload"/>}
```
maxbates: is this just for testing?

src/components/modal/modalwindow.js
duncanmeech: const contents is the actual window ( versus the modal block which is always rendered on or off screen ). The contents should not be rendered when the modal isn't open.

src/containers/graphics/views/constructviewer.js
duncanmeech: Yes, but since there are no uses of the modal window yet it will have to stay until we have something else to test.

src/components/modal/modalwindow.js
duncanmeech: Well not perfect but they make sense to me? In the context of a modal window class things like title ( string ), open ( boolean ) seem ok.

PR #229, opened 2016-01-29
Onion as node module

dchandran:
Modules can be local or from npm repository. When they are local, registeredExtensions.json will contain a local path. Currently, the examples are local - see registeredExtensions.json.

How to make Onion work in this new framework: 
```
cd extensions
git clone https://github.com/IsaacLuo/onion2.git
```
Once we know all the info that registeredExtensions.json needs to contain, we can add more fields.

@IsaacLuo would need to merge my pull request in the onion repo - I added package.json and index.js


extensions/requireExtensions.js
```diff
@@ -1,17 +1,38 @@
...
+        exports.extensionsInfo[namespace] = {};
+      }
+      if (!(namespace in exports.errors)) {
+        exports.errors[namespace] = {};
+      }
+      for (name in extensions) {
+        try {
+          path = extensions[name].path;
+          exports.extensions[namespace][name] = require(path);
+          exports.extensionsInfo[namespace][name] = require(path + '/package.json');
```
maxbates: This is not likely to be compatible with npm... what information are you expecting here?

maxbates:
This looks fine, but migration to npm may not be entirely clear. We should not put strict requirements on structure of an npm module's `package.json` -- lets be clear about what data we expect, and can require an additional field on the file with our custom metadata

extensions/requireExtensions.js
dchandran: At some point, we might have a GUI for viewing, downloading, and removing extensions. I was hoping that the package.json can provide the information for the GUI.

maxbates:
can you merge in master (removes authentication middleware) then good to go

maxbates:
THis work has been merged into branch `npm-extensions` and continued there

PR #232, opened 2016-02-03
-remove redis requirement for app

duncanmeech:
-remove authentication, soon to be replaced.
-leave only working stubs in database.js that always return resolved promises


maxbates:
need to remove from dockerfile, otherwise its more or less OK since I'll be changing / replacing all this myself

maxbates:
looks already done. merging

PR #235, opened 2016-02-06
Inspector multiselect

duncanmeech:
Adds multi-select for the inspector. Has ghost inspector test separately.
Would be nice to get in for Monday Demo @maxbates


maxbates:
Probably want to change which fields are visible, but fine for now

PR #239, opened 2016-02-11
update to npm3 - requires local update

maxbates:
Everyone will need to install npm3 locally after merging. Realistically, this doesnt cause any changes yet, except to speed things up a bit, as dependencies are now installed maximally flat

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ ADD . /app
 
 #setup node
 ADD package.json /app/package.json
-RUN npm update -g npm
+RUN npm install -g npm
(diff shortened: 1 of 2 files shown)


duncanmeech:
So to install npm3 with conflicting with other npm projects we would do:
npm install -g npm3
And then we would need to enter things like 'npm3 checkout -b new-branch' to ensure we are using the correct version for the project?

maxbates:
I am confused what the last sentence means?
On a cursory search didnt see an easy to way only install npm3 for a specific project.

PR #240, opened 2016-02-12
Persistence - commits + snapshots

maxbates:
Updates to the persistence layer and REST API
Adds a /:projectId/commit route to create snapshots
All CRUD operations create commits
Adds commitMessage module for future parsing / filtering of commit messages
Move findProjectFromBlock.js to new file

includes some updates to blockGetSequence() but not completed (other work will build on this in the future)
deprecate several tests (move to new folder, outside of Mocha scope)


duncanmeech:
Needs a rebase and/or merge before can be accepted. Also, I'd like to walk through this one with you.

PR #249, opened 2016-02-22
Nested constructs

duncanmeech:
Nested constructs. Layout / connectors will be updated in subsequent PR.


maxbates:
+1 once merged

PR #250, opened 2016-02-23
Parent schema update

maxbates:
`parent` schema update to form `{id: <id>, sha: <sha>}` where SHA is a git SHA, the version for the project, or the project version for a block.

Update blockClone() to require projectVersion as argument, add selector to get projectVersion, update constructviewer.js cloning. Update tests.


duncanmeech:
:+1:

PR #252, opened 2016-02-24
make 'npm install' compatible with Windows

IsaacLuo:
currently, 'npm install' in windows will fail, because the js file can only be run by the command 'node xxx.js'


maxbates:
already done

PR #261, opened 2016-03-10
Grunt fix

duncanmeech:
Move grunt to global nav, make visible on homepage, add test


maxbates:
pending auth branch, will check then

PR #265, opened 2016-03-12
basic md5->sequence caching using local storage

maxbates:


src/middleware/localStorageCache.js
```diff
@@ -0,0 +1,14 @@
...
+
+export function getItem(id) {
+  return storage.getItem(id);
+}
+
+export function setItem(id, value) {
+  invariant(id, 'id is required to setItem');
+  invariant(typeof value === 'string', 'value must be a string');
+
+  return storage.setItem(id, value);
```
duncanmeech: This can throw exceptions, would that break the app?

PR #269, opened 2016-03-18
fix bad command spec and add ENV variable for auth API end-point

dhylbert:
This fixes the command spec for the quick-start configuration and defines the Authentication REST API endpoint via an environment variable.


dhylbert:
This needs to be merged into master to keep the Jenkins build job running successfully.

PR #272, opened 2016-03-21
Various Fixes for EGFAD-102

duncanmeech:
- Fixes exception when dropping an sbol block on an empty construczt
- Optimizations for rendering constructs including debouncing the trailing edge of calls to update, because react will send several renders for something like a drop operation i.e. blockCreate, blockClone, selectConstruct, addBlock all cause updates.
- give emphasis to construct selection by lightening the selected construct and darkening others. 
- range selection is enabled for the same level in the construct ( but not ordering of selections ).
- show sbol symbol on sbol symbols dragged from inventory.
- moves udpateAccount to middleware which fixes the form and associated test.

Other items in the card are addressed in upcoming cards.


maxbates:
Does this include updating the E2E tests to accommodate the new inventory tab?

PR #274, opened 2016-03-22
fetch rejects on 4xx and 5xx in middleware

maxbates:


duncanmeech:
Signout is broken and possibly other authentication XHR requests.
When npm run nightwatch passes all tests we can take another look ( you might want to pre-merge this with master as it has a fix for the update account form and associated test )

PR #276, opened 2016-03-22
fix removeEventListener to window resize event in construct viewer

duncanmeech:


maxbates:
good to go after you merge other pulls :)

PR #284, opened 2016-03-25
Inventory Search - tabbed by type / by source

maxbates:
(also includes transactions pull request)

Support in search inventory to sort by source or by kind. It's a little ugly, but will clean it up when have to handle in the My Projects tab as well (unsure what generalizes)


duncanmeech:
@maxbates The UX is a bit flaky i.e. search for 'cds' ...results appear, then open another inventory section, then go back to search and the results are gone....also hitting enter on the search box won't perform another search till you change the text.
Also, you should seriously consider adding some e2e tests for this UI stuff. It would be really easy to setup something to test this entire branch. I'll help you when back.

PR #286, opened 2016-03-25
Project searching

maxbates:
basic querying for all blocks, blocks by type, blocks by name. May move to a grep if need more speed, short of using a database.

adds `Block.classless` and `Project.classless` static methods to create unfrozen objects without instance methods. Also adds `Block.validate` static method, and project too.


duncanmeech:
Doesn't appear to add any search functionality to the UI as yet...but the tests all pass!

PR #293, opened 2016-03-29
use private DNS for Authentication API HTTP endpoint

dhylbert:
This change isn't urgent, but will be necessary when User Authentication is required for QA and Production environments.

`auth.bionano.bio` and `auth.dev.bionano.bio` point to the same internal IP address in AWS and are private DNS records available only in Bionano AWS VPC Networks.


duncanmeech:
:+1:

PR #298, opened 2016-04-04
Export Genbank

florencio-mazzoldi:
Cleaned up the code.
Split import and export python files
Export should work now
Added tests


plugins/convert/genbank/genbank_export.py
```diff
@@ -0,0 +1,119 @@
+import json
```
duncanmeech: Looks like you accidentally removed all the comments from this code. What does it do?

PR #301, opened 2016-04-05
Copy paste enhance

duncanmeech:
Card has story: https://egf-atgc.atlassian.net/secure/RapidBoard.jspa?rapidView=1&view=detail&selectedIssue=EGFAD-289


src/containers/GlobalNav.js
```diff
@@ -32,6 +32,9 @@ import { undo, redo } from '../store/undo/actions';
 import { uiShowGenBankImport } from '../actions/ui';
 import { setItem } from '../middleware/localStorageCache';
 import { uiShowDNAImport } from '../actions/ui';
+import {
+  copyBlocksToClipboard
```
maxbates: could this be an action?

src/containers/uiapi.js
```diff
@@ -0,0 +1,35 @@
+import { clipboardSetData } from '../actions/clipboard';
```
maxbates: let's get containers / components folders consistent. This isn't a component and should go somewhere else.
Idea between the two was that containers were linked to the store, components were purely driven by properties. Maybe we should pst-fix containers e.g. Inventory-container.js... but anyway, this should go into another folder. Why aren't these just actions?

duncanmeech: moved to utils/ui

src/containers/GlobalNav.js
duncanmeech: Just an experiment that didn't work out. It is an action, that method was not used.

PR #306, opened 2016-04-07
Egfad 32

duncanmeech:


src/containers/GlobalNav.js
```diff
@@ -53,6 +68,93 @@ class GlobalNav extends Component {
...
+  onSelectAll() {
+    this.props.focusBlocks(this.props.blockGetChildrenRecursive(this.props.focus.construct).map(block => block.id));
+  }
+
+  /**
+   * save current project and signal this as the most recent project to reopen
+   */
+  saveProject() {
+    this.props.projectSave(this.props.currentProjectId);
+    setItem('mostRecentProject', this.props.currentProjectId);
```
maxbates: This should go in the action so it is applied consistently

src/actions/projects.js
```diff
@@ -36,6 +38,7 @@ export const projectCreate = (initialModel) => {
 
 //Promise
 export const projectSave = (projectId) => {
+  setItem('mostRecentProject', projectId);
```
maxbates: can you put this after L43 (get the project) to make sure it exists

PR #308, opened 2016-04-07
Fixed tests

florencio-mazzoldi:
Several issues on testing. Added error logging for python errors. 

The tests don't run reliably 100% of the time yet. Jira ticket opened for Max to investigate a User ID issue with the REST API.


florencio-mazzoldi:
Now it also doesn't crash for weird blocks. Max, Insulin works. However, there is something to discuss about annotations in the root blocks (constructs). They will never show! That's what happens with the misc_feature block in HUMINS02 in the file you sent me.

maxbates:
This is because they dont have any sequence? We may show them in the inspector, but if they are sequence annotations then this is correct, they will not show up anywhere (but they shouldnt be on a block without a sequence then)

florencio-mazzoldi:
If the only outstanding issue is the fact that we are using Annotation for the "parent" blocks, can you please approve this PR? I'll create a ticket in Jira to move from using Annotations to Notes in non-leaf blocks.

PR #309, opened 2016-04-08
Ui updates

maxbates:


maxbates:
note duncan - consolidates colors

PR #318, opened 2016-04-11
Autosaving

maxbates:
EGFAD 166


maxbates:
updated

duncanmeech:
not autosaving...its that bad? ;)

maxbates:
could use some debouncing for initiating saves, but requires rewriting some tests and not enormously important yet. but should make card to come back to timing of this.

duncanmeech:
I see this in the console when I run this branch: Response {type: "basic", url: "http://localhost:3001/data/projects/project-2c0341f7-3dc0-4d97-b2e5-ca09d4bc33aa", status: 400, ok: false, statusText: "Bad Request"…}

PR #323, opened 2016-04-12
Genbank

florencio-mazzoldi:


plugins/convert/export.js
```diff
@@ -33,8 +35,12 @@ function callExportFunction(funcName, field, id, input) {
...
 }
 
-function exportProject(id, input) {
-  return callExportFunction('exportProject', 'project', id, input);
+function exportProject(id, roll) {
+  return callExportFunction('exportProject', 'project', id, roll);
+}
+
+function exportConstruct(id, roll, constructId) {
+  return callExportFunction('exportConstruct', 'project', id, { roll: roll, constructId: constructId });
```
duncanmeech: when property names match your variables you can do this with:
{roll, constructId}

PR #333, opened 2016-04-14
server reloads on changes, doesnt reload client

maxbates:


maxbates:
@duncanmeech make sure that change made locally is overridden with this one

maxbates:
wait a moment on this actually
L113 of start needs to be changed to a noop

PR #337, opened 2016-04-15
Querying components

maxbates:
Support on Data API to get components of a block
Dragging constructs from inventory should work fine now. 
Future support for when we don't load the whole project into the store on toggling inventory tabs


maxbates:
needs a test, coming!

maxbates:
ready @duncanmeech

PR #341, opened 2016-04-18
Block focus fix

duncanmeech:


duncanmeech:
JIRA card here: https://egf-atgc.atlassian.net/browse/EGFAD-321

PR #343, opened 2016-04-20
Autoscroll

duncanmeech:
https://egf-atgc.atlassian.net/browse/EGFAD-333


maxbates:
should switch to lodash, but we should just get rid of that whole utils folder and swap it out for lodash. just make sure they work they same so that switch is trivial. otherwise good to go

PR #344, opened 2016-04-20
Egfad 328

duncanmeech:
reset password now does login after reset, except that the login method wasn't required on the page, now it is.


duncanmeech:
p.s. this can really only be tested in production since locally the reset password request doesn't generate the required email.

maxbates:
+1

PR #345, opened 2016-04-20
Egfad 147

duncanmeech:


maxbates:
Does this handle horizontal scroll bars?

maxbates:
otherwise good

PR #346, opened 2016-04-21
Bug squashes

maxbates:
EGFAD 332
EGFAD 285


duncanmeech:
When I select 'New Project' from the file menu I get client.js:11605 Uncaught TypeError: this.isFiller is not a function
Nightwatch tests are all failing when they try to create a new project

PR #347, opened 2016-04-21
Express errors

maxbates:
EGFAD 312


maxbates:
missing a commit, github not picking it up, see #348

PR #349, opened 2016-04-21
- Focus can be switched between blocks, construct ( by clicking aways…

duncanmeech:
Focus can be switched between blocks, construct ( by clicking away from blocks or construct title ) and project ( by clicking project header ).

- End to end tests for focus / name changes for blocks, constructs, projects.
- Decorate blocks in the construct viewer with optional data-??? attributes for easier testing

### DIFF
--- a/src/containers/Inspector.js
+++ b/src/containers/Inspector.js
@@ -27,6 +27,11 @@ export class Inspector extends Component {
   render() {
     const { isVisible, instances, project, readOnly } = this.props;
 
+    // inspect instances, or construct if no instance or project if no construct or instances
+    let inspect = instances && instances.length
+    ? <InspectorBlock instances={instances} readOnly={readOnly}/>
+  : this.props.constructId ? <InspectorBlock instances={[this.props.blocks[this.props.constructId]]} readOnly={readOnly} />: <InspectorProject instance={project} readOnly={readOnly}/>;
+
     return (
       <div className={'SidePanel Inspector no-vertical-scroll' +
       (isVisible ? ' visible' : '') +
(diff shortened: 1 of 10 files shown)


src/containers/Inspector.js
```diff
@@ -81,10 +82,13 @@ function mapStateToProps(state, props) {
     isVisible,
     readOnly,
     instances,
+    constructId : construct ,
     project,
+    blocks: state.blocks,
```
maxbates: This will slow things down... components should only get what they need from the store. this is perhaps why your viewers update all the time -- any single change that affects the block store is now going to re-render this component

src/containers/Inspector.js
```diff
@@ -27,6 +27,11 @@ export class Inspector extends Component {
   render() {
     const { isVisible, instances, project, readOnly } = this.props;
 
+    // inspect instances, or construct if no instance or project if no construct or instances
+    let inspect = instances && instances.length
+    ? <InspectorBlock instances={instances} readOnly={readOnly}/>
```
maxbates: No need to complicate this, add this logic to the mapStateToProps, and just pass the construct as the instance

src/containers/graphics/views/layout.js
```diff
@@ -276,7 +278,7 @@ export default class Layout {
 
       // update title to current position and text
       this.titleNode.set({
-        text: this.construct.metadata.name || this.construct.id,
+        text: this.construct.metadata.name || 'Construct',
```
maxbates: I believe it's "New Construct"

PR #350, opened 2016-04-21
Egfad 322

duncanmeech:
Construct color from metadata and not random any more.


maxbates:
looks good, pending issues in previous pull

PR #354, opened 2016-04-22
Egfad 324

duncanmeech:
Update import DNA as per spec in JIRA card ( https://egf-atgc.atlassian.net/browse/EGFAD-324 )

NOTE: Also extends allowable DNA chars to:  # acgturyswkmbdhvn.-


maxbates:
@IsaacLuo does onion allow degenerate bases?
@florencio-mazzoldi does genbank import/export allow degenerates?

florencio-mazzoldi:
Genbank allows it because BioPython allows it.

maxbates:
cool

PR #363, opened 2016-04-26
Ncbi inventory

maxbates:


duncanmeech:
Exceptions on empty results set and other conditions? Needs better error handling.

maxbates:
@duncanmeech ok

PR #368, opened 2016-04-27
Block nesting and push down

duncanmeech:
Primarily EGFAD-393. Blocks, including EG parts and sketch blocks will now nest if dropped on the center of the target block.
Also EGFAD-342, issues with connectors not rendered sometimes.
Also EGFAD-200 when a blocked on a parent block that already has DNA then a new block is created to replace the parent and the parent AND the drop payload are added as children ( or at the end of existing children ).


src/actions/blocks.js
```diff
@@ -247,6 +250,24 @@ export const blockSetSbol = (blockId, sbol) => {
...
+ * Create a new child block and set the given sbol symbol
+ */
+export const blockAddSbol = (blockId, sbol) => {
+  return (dispatch, getState) => {
+    dispatch(undoActions.transact());
+    const oldBlock = getState().blocks[blockId];
+    const newBlock = dispatch(blockCreate());
+    dispatch(blockSetSbol(newBlock.id, sbol))
+    dispatch(blockAddComponent(oldBlock.id, newBlock.id, oldBlock.components.length));
+    //end transaction
```
maxbates: you never actually dispatch the event BLOCK_ADD_SBOL ... so you dont need to make a new action for it and can remove it from the reducer

maxbates:
remove constant too and ready to go

duncanmeech:
Done...merging

PR #370, opened 2016-04-28
Sonic the hedgehog

duncanmeech:
- Construct viewer header ( triangle ) changes color with construct

- double click a block to open the inspector on that block


src/containers/graphics/views/constructviewer.js
```diff
@@ -334,6 +358,13 @@ export class ConstructViewer extends Component {
     this.setState(state);
   }
   /**
+   * open the inspector
+   * @return {[type]} [description]
+   */
+  openInspector() {
```
maxbates: why more abstraction over actions?

maxbates:
ok after changes to prior pull. please branch off master.

src/containers/graphics/views/constructviewer.js
duncanmeech: constructviewer is a component but it owns a non react component that handles all the user interface interaction. Easier to call methods on the component than to dispatch actions from non react component.

PR #371, opened 2016-04-28
Project ID in block + block.parents

maxbates:
EGFAD 404
EGFAD 311

store projectId on blocks, update schemas etc.

store projectId in block.parents so you can actually track the history of a block (without looking up the project dynamically at every single step)

update blockCreate and blockClone methods + usage


duncanmeech:
needs a merge

PR #373, opened 2016-04-28
Help Menu / Discourse / About Dialog

duncanmeech:
add discourse base URL from ENV_VAR and flash to client. Add about dialog. Connect all items on the help menu


maxbates:
If this is how we are handling the about page, there is a route for the about page that should be removed. see /routes/

duncanmeech:
About page remove, merging.

PR #375, opened 2016-04-29
save project before attempting to download genbank file

duncanmeech:
egfad-377

### DIFF
--- a/src/containers/GlobalNav.js
+++ b/src/containers/GlobalNav.js
@@ -145,10 +145,10 @@ class GlobalNav extends Component {
   }
 
   /**
-   * save current project and signal this as the most recent project to reopen
+   * save current project, return promise for chaining
    */
   saveProject() {
-    this.props.projectSave(this.props.currentProjectId);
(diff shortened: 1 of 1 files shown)


src/containers/GlobalNav.js
```diff
@@ -182,12 +182,15 @@ class GlobalNav extends Component {
...
   downloadProjectGenbank() {
-    // for now use an iframe otherwise any errors will corrupt the page
-    const url = `${window.location.protocol}//${window.location.host}/export/genbank/${this.props.currentProjectId}`;
-    var iframe = document.createElement("iframe");
-    iframe.style.display = "none";
-    iframe.src = url;
-    document.body.appendChild(iframe);
+    this.saveProject()
+      .then(() => {
+        // for now use an iframe otherwise any errors will corrupt the page
```
maxbates: what errors?

duncanmeech: Errors in the genbank generation code will produce JSON which the browser replaces the page with. This is a weakness of this design. Using an iframe prevents that.

PR #376, opened 2016-04-29
Egfad 419

duncanmeech:
Fix all end to end tests so they don't rely on the old test project.


maxbates:
will review once other branches are merged

PR #381, opened 2016-05-03
Csv upload

duncanmeech:
Switch CSV and GB uploads to formidable file handling ( which is already a part of Express ).
Going forward putting text in the body of the request won't work. You can still use Postman, but use a file upload.
Also adds real tests for CSV and GB uploads into new project and GB download.


duncanmeech:
JIRA: https://egf-atgc.atlassian.net/browse/EGFAD-405

plugins/convert/csv/index.js
```diff
@@ -131,15 +132,14 @@ const handleProject = (outputProject, rootBlockIds) => {
   }));
 };
 
-// Reads a cvs file and returns a project structure and all the blocks
+// Reads a csv file and returns a project structure and all the blocks
 // These return structures are NOT in GD format.
-const readCvsFile = (cvsString) => {
```
florencio-mazzoldi: Wow that typo was all over the file. Thank you!

plugins/convert/genbank/index.js
```diff
@@ -135,7 +135,6 @@ const handleProject = (outputProject, rootBlockIds) => {
 // These return structures are NOT in GD format.
 const readGenbankFile = (genbankString) => {
   const inputFile = createRandomStorageFile();
```
florencio-mazzoldi: Shouldn't we point the python routine directly to the file downloaded by formidable? That would remove an unnecessary copy...

florencio-mazzoldi:
E2E Tests pass. As soon as that "copy" comment is addressed we can merge it.

plugins/convert/genbank/index.js
duncanmeech: looks like CSV and Genbank could both just use the file path but the linkage to the plugins was not clear to me so I didn't touch it. I'd rather add a backlog card for that task and address later ( plus Max has a hack in there about passing strings directly in which would need touching as well )

duncanmeech:
Also, you might want to try one more merge with master. I just merged a branch of Max's that has broken e2e. Want to make sure it doesn't merge badly with this PR

florencio-mazzoldi:
Ok, please add the card. I'll re-merge with master and test again.

plugins/convert/import.js
```diff
@@ -42,53 +44,53 @@ function importProject(pluginId, data) {
...
 //can pass :projectId = 'convert' to just convert genbank file directly, and not write to memory
 router.post('/:pluginId/:projectId?', jsonParser, (req, resp, next) => {
-  const { pluginId, projectId } = req.params;
+  const {pluginId, projectId} = req.params;
 
-  //assuming contents to be string
-  let buffer = '';
+  // needed for genbank without a project.
+  if (projectId === 'convert') {
+    return callImportFunction('convert', pluginId, data)
```
duncanmeech: Good catch, I'll fix that but I think the test is all you. Let me know if you need help.

PR #383, opened 2016-05-03
fixes issues with empty search results from various sources

duncanmeech:
EGFAD-439


maxbates:
this actually doesn't - it will show 0 sometimes and format wrong. this is handled in my pull request

PR #387, opened 2016-05-05
Egfad 428

duncanmeech:
Leave filler blocks with no name.
Ensure inspector can handle blocks with no name or symbol


duncanmeech:
JIRA: https://egf-atgc.atlassian.net/browse/EGFAD-428

PR #389, opened 2016-05-05
Inventory hierarchy

maxbates:
Projects + Constructs appear in inventory properly. Some style + code clean up aronud that as well.


duncanmeech:
It had same issues as another branch of yours but was fixed by merging with master.
However, I also see this warning when the CSS is transpiled:

postcss-custom-properties: /Users/meechd/dev/genome-designer/src/styles/Toggler.css:12:3: variable '--timing-extrafast' is undefined and used without a fallback

Also a lot of the e2e tests are failing e.g. block.getName() is not defined and others possibly due to inventory changes. We will have to review tomorrow.

maxbates:
dealt with variables issue
cant test end to end..

maxbates:
Need to remove `block.metadata.isSbol` in `InventoryItemSbol`, or ensure that the fieldi s removed when the block is added to the canvas. on @duncanmeech

PR #390, opened 2016-05-06
only flatten dropped constructs when dropping into a new construct

duncanmeech:
https://egf-atgc.atlassian.net/browse/EGFAD-445


src/containers/graphics/views/constructviewer.js
```diff
@@ -219,7 +219,8 @@ export class ConstructViewer extends Component {
 
     // if the source is the inventory and we are dragging a single block with components
     // then we don't want to insert the parent, so replace the payload with just the children
-    if (!Array.isArray(payload.item) && payload.source === 'inventory' && payload.item.components.length ) {
+    // NOTE: Flatten only for new constructs
+    if (newConstruct && !Array.isArray(payload.item) && payload.source === 'inventory' && payload.item.components.length ) {
```
maxbates: Weird naming but does behave correctly, minus note in comments

maxbates:
This does not accommodate dragging a construct from a construct in the canvas into the "drop blocks here" section. If you drag a construct from an existing construct onto that box, you get an additional parent. I would expect that dragging a construct onto that box makes the parent the construct, not a child of a new one.

duncanmeech:
Yes, I missed that case. I'll will also flatten when adding to an empty construct, not just a new one.

duncanmeech:
@maxbates updated as discussed and tests added for dragging construct and blocks from a saved project.

PR #391, opened 2016-05-07
Performance - ability to pause store updates

maxbates:
Adds a store enhancer so you can pause the store from updating subscriptions, and allows you to dispatch actions to pause / resume. Should have no other side effects.

Applies these actions in some costly operations e.g. loading projects, cloning many blocks, etc.


duncanmeech:
Noticeably faster when dragging / dropping large block sequences!

PR #394, opened 2016-05-08
Lint redux

duncanmeech:
Lint the files I've touched since the last time I linted....now it's your turn.


src/components/modal/modalwindow.js
```diff
@@ -1,6 +1,5 @@
 import React, { Component, PropTypes } from 'react';
 import ReactDOM from 'react-dom';
-import ReactTransitionGroup from 'react-addons-css-transition-group';
```
maxbates: remove from package.json?

maxbates:
see ccomments, cool otherwise

src/components/modal/modalwindow.js
duncanmeech: indeed, it was removed from there as well.

PR #402, opened 2016-05-11
Triangles

maxbates:


src/components/ui/Toggler.js
```diff
@@ -14,21 +19,39 @@ export default function Toggler({ onClick, hidden, open, disabled }) {
...
 
-  return (<span className={'Toggler' +
+  const templateId = `disclosure_triangle_closed`;
+  const template = document.getElementById(templateId);
+  const svg = template.cloneNode(true);
+  svg.removeAttribute('id');
+
+  const markup = serializer.serializeToString(svg);
+
+  return (<div className={'Toggler' +
```
duncanmeech: nested ternaries....lint will not like this.

PR #405, opened 2016-05-11
Added hierarchical information to the genbank file as it exports

florencio-mazzoldi:
Updated the ID of the constructs as they are read from genbank. Fixed tests.

@maxbates there are 2 tests on Matrix2D that fail... please fix them.

@Flo, fixed in master, merge with master to fix.


maxbates:
those are duncans tests

plugins/convert/genbank/index.js
```diff
@@ -226,16 +227,14 @@ const exportProjectStructure = (project, blocks) => {
...
-    })
+    .then(resStr => Promise.resolve(resStr))
     .catch(err => {
       console.log('ERROR IN PYTHON');
       console.log('Command');
       console.log(cmd);
       console.log('Error')
       console.log(err);
-      return Promise.reject(err);
+      Promise.reject(err);
```
maxbates: you should return this or it will be swallowed

plugins/convert/genbank/index.js
```diff
@@ -58,7 +58,8 @@ const createBlockStructureAndSaveSequence = (block) => {
       annotations: block.sequence.annotations, //you will likely have to remap these...
     },
     source: {
-      id: 'genbank',
+      id: block.version ? block.version : '',
```
maxbates: This should be the name of the file, which is written in a consistent location and not temporary. See https://egf-atgc.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=EGFAD&view=detail&selectedIssue=EGFAD-505

maxbates: The idea is that the file will then be retrievable. Ideally, if there is an ID in a genbank file, you would use that, and the file name is set to the same ID (so that it can be fetched later simply by the ID)

duncanmeech:
@florencio-mazzoldi In addition to Max's comments also needs a rebase now as well.

florencio-mazzoldi:
@duncanmeech @maxbates Fixed the import/export tests. CSV + Genbank work well now. All tests passed.

plugins/convert/genbank/genbank_import.py
```diff
@@ -96,6 +96,7 @@ def create_root_block_from_genbank(gb, sequence):
     root_block["metadata"]["end"] = full_length - 1
     root_block["metadata"]["genbank"]["id"] = gb.id
     root_block["sequence"]["length"] = full_length
+    root_block["version"] = gb.id
```
maxbates: what is this?

plugins/convert/genbank/genbank_import.py
```diff
@@ -291,6 +308,12 @@ def create_filler_blocks_for_holes(all_blocks, sequence):
             all_blocks[block_id] = filler_block
             block["components"].insert(i + 1, block_id)
 
+# If a block has children, remove its sequence
+def remove_sequence_from_parents(all_blocks):
+    for block in all_blocks.values():
+        if len(block["components"]) > 0:
+            block["sequence"]["sequence"] = ""
+            block["sequence"]["length"] = 0
```
maxbates: this should probably be null, not 0. not a big deal.

plugins/convert/genbank/index.js
```diff
@@ -18,33 +18,33 @@ const uuid = require('node-uuid');
...
-        });
-      });
-    })
-    .then(() => fileSystem.fileRead(outputFile, false));
+const runCommand = (command, inputFile, outputFile) => {
+  return new Promise((resolve, reject) => {
+    exec(command, (err, stdout) => {
+      if (err) {
+        console.log('ERROR!!!!!');
+        console.log(err);
```
maxbates: might be more helpful to log `err.stack` ? not sure, depends on what youre seeing

plugins/convert/genbank/index.js
```diff
@@ -1,14 +1,14 @@
-import * as filePaths from '../../../server/utils/filePaths';
 import * as fileSystem from '../../../server/utils/fileSystem';
```
maxbates: I feel like I'm having deja vu reviewing this file....

plugins/convert/import.js
```diff
@@ -53,51 +57,61 @@ router.post('/:pluginId/:projectId?', jsonParser, (req, resp, next) => {
       buffer += data;
     });
     req.on('end', () => {
-      return callImportFunction('convert', pluginId, buffer)
-        .then(converted => resp.status(200).json(converted))
-        .catch(err => next(err));
+      const tempFilePath = filePaths.createStorageUrl('temp/' + uuid.v4());
```
maxbates: Will this still preserve the genbank file and add it to the block source? It should.

florencio-mazzoldi: @maxbates I thought that for NCBI we didn't want to preserve the file... the idea would be to link to the NCBI version of it, is it not?

plugins/convert/genbank/index.js
florencio-mazzoldi: It was weird, it felt deja-vu adding the paths in... somehow it disappeared.

plugins/convert/genbank/index.js
florencio-mazzoldi: I kept the console.logs in the case of errors separately than the return back... it was helpful to see those while debugging.

plugins/convert/genbank/genbank_import.py
florencio-mazzoldi: Will remove.

plugins/convert/import.js
florencio-mazzoldi: Ok I saved the file there as well.

plugins/convert/import.js
```diff
@@ -57,60 +57,59 @@ router.post('/:pluginId/:projectId?', jsonParser, (req, resp, next) => {
...
-              });
-          });
-      });
+    // save incoming file then read back the string data.
+    // If these files turn out to be large we could modify the import functions to take
+    // file names instead but for now, in memory is fine.
+    const form = new formidable.IncomingForm();
+    form.parse(req, (err, fields, files) => {
+      fileSystem.fileRead(files.data.path, false)
+        .then((data) => {
```
maxbates: The nice thing about promises is you dont get pyramid code because you can chain them 👎  this will work but youre showing off your python :)

maxbates:
Do you ever set the md5 on `source.id`? How am i supposed to look it up?

plugins/convert/import.js
```diff
@@ -57,60 +57,59 @@ router.post('/:pluginId/:projectId?', jsonParser, (req, resp, next) => {
...
-      fs.writeFile(tempFilePath, buffer, (err) => {
-        return callImportFunction('convert', pluginId, tempFilePath)
-          .then(converted => {
-            fs.unlink(tempFilePath, (err) => {
-              resp.status(200).json(converted);
-            });
-          })
-          .catch(err => next(err));
-      });
+      const inputFilePath = filePaths.createStorageUrl(pluginId + '/' + md5(buffer));
```
maxbates: For genbank conversion, you may want to just do this in the genbank plugin directly, and have the plugin set the source.source and source.id ,rather than trying to make it generic. not sure all plugins will want to do this? or just leave it. Maybe they do. but need `source.id` on the block set somewhere

florencio-mazzoldi:
md5 stored in the source.id (line 61 of genbank/index.js). Note that I only store the file name, without the rest of the path (line 47), assuming you can get to it.

PR #407, opened 2016-05-11
create storage directories on container build

maxbates:


Dockerfile
```diff
@@ -26,6 +26,11 @@ EXPOSE 3000
 ENV PORT=3000
 
 RUN mkdir /app
+#do some directory setup for now, but the server should handle this...
+RUN mkdir /app/storage
```
duncanmeech: are the default permissions that are reliable here? Or should we chmod these folders just to be safe?

maxbates: @PeterRJones ?

PR #408, opened 2016-05-11
huge refactor from rules.sbol to rules.role, rename SBOL specific com…

maxbates:
…ponents and fields to role

it's a juicy one!


maxbates:
If not a problem, just look at #409 and merge htat one, so dont have to deal with merge again

PR #417, opened 2016-05-13
We finally get a database

duncanmeech:
Initial Mongo DB integration with updates to readme and a simple route for creating and saving an order.


server/orders/orders.js
```diff
@@ -0,0 +1,35 @@
...
+  limit: '50mb',
+});
+
+/***************************
+ Middleware
+ ****************************/
+
+router.use(jsonParser);
+router.use(errorHandlingMiddleware);
+// ==================================================================
```
maxbates: This likely should communicate with the other parts of the server where project is saved and just fetch it from there? This routing is all stub, I get it

server/orders/orders.js
```diff
@@ -0,0 +1,35 @@
+import express from 'express';
```
maxbates: should be index.js

server/moogoose-models/orders.js
```diff
@@ -0,0 +1,72 @@
+import mongoose from 'mongoose';
+const Schema = mongoose.Schema;
+const ObjectId = mongoose.Schema.Types.ObjectId;
+import uuid from 'node-uuid';
+
+/**
+ * basic order schema
+ * @type {mongoose.Schema}
+ */
+var orderSchema = new Schema({
```
maxbates: will come from /src/schemas/ in the future...

maxbates:
not going to merge this because it looks like it may cause problems on the build? And no rush?

PR #418, opened 2016-05-13
Verify projectid

maxbates:


duncanmeech:
@jlachoff lets discuss the never flattening constructs with Joe. Also a lot of the e2e tests are broken here and in master due to your ribbon changes I suspect.

maxbates:
ribbon changes were reversed in one of the ones you just merged

duncanmeech:
In that case I've not idea....opening the inventory fails in all the test because the trigger is hidden behind the ribbon I think.

maxbates:
ah yes, the ribbon is in the way, I am guessing it may be because that phantom div you had (that was pushing the design canvas down) was allowing it to be seen when there was a ribbon

duncanmeech:
possible, but that was only in the upload genbank dialog, not normally seen.

maxbates:
Anything you try to open a project that doesnt exist it also shows up

jlachoff:
let me know if you still need to talk with me about this - I have no idea what it's about...

maxbates:
@jlachoff beofre there was "flattening" of the construct when you dropped it - its components were added rather than a single block with components in it. I dont think we ever want that to happen. When you drag-and-drop, dropping blocks is simple, dropping constructs should always add a single block with its children still its children. The only special handling, visually, is when you drop a construct into the project vs. into another construct -- dropping a construct into a construct vs. creating a construct in that project. Never do you want to just add the components without the parent itself, right?

maxbates:
Ok @duncanmeech this now includes the grunt fixes  #422 (grunt fix) and #421 (empty components) and #424 (file fixes). Tests should pass.

PR #432, opened 2016-05-17
constructs default to name 'new construct'

maxbates:


duncanmeech:
Approved one the branch is rebased.

PR #440, opened 2016-05-19
Egfparts listblocks

maxbates:
includes #439 and #426 

creates testing project at /project/combinatorial with a template and list block. Inspector has basic list showing when you click the list block (interaction all wrong with that part though)


duncanmeech:
needs a rebase

PR #451, opened 2016-05-20
Template layouts

duncanmeech:
for /projects/combinatorial this implements list blocks rendering and the basics of editing constraints around frozen/fixed constructs


src/containers/graphics/views/layout.js
```diff
@@ -12,6 +13,1962 @@ import invariant from 'invariant';
 const blockType = 'block';
 const roleType = 'role';
 
+const listStore = [
```
maxbates: oh god no. just import parts.js - the IDs will change soon anyway.

maxbates: but added a commit last week that puts them all in the store so you shouldnt need this anwyay

src/models/Instance.js
```diff
@@ -24,7 +24,8 @@ export default class Instance {
     );
 
     if (process.env.NODE_ENV !== 'production') {
-      require('deep-freeze')(this);
+      console.warn();('Objects are not frozen for temporary testing.');
+      //require('deep-freeze')(this);
```
maxbates: Why do you need this?

src/containers/graphics/views/layout.js
duncanmeech: Its just temporary...so you can ignore. I don't really want to keep PR's in a perpetual state of update while we finalize functionality. Why not merge this one and adjust in the inevitable next PR?

src/models/Instance.js
duncanmeech: Was necessary when I was hacking in list blocks for testing. Needs to be removed, next commit.

src/containers/graphics/views/layout.js
```diff
@@ -3,15 +3,23 @@ import Vector2D from '../geometry/vector2d';
...
 import kT from './layoutconstants';
 import objectValues from '../../../utils/object/values';
 import invariant from 'invariant';
+import partList from '../../../inventory/andrea/partList.json';
 
 // just for internal tracking of what type of block a node represents.
 const blockType = 'block';
 const roleType = 'role';
 
+// map the parts to a hash by part ID
```
maxbates: they are in the store now...

src/containers/graphics/views/layout.js
```diff
@@ -514,12 +564,17 @@ export default class Layout {
...
     let nestedVertical = 0;
 
+    // additional height required by the tallest list on the row
+    let maxListHeight = 0;
+
     // width of first row is effected by parent block, so we have to track
     // which row we are on.
     let rowIndex = 0;
+    // display only non hidden blocks
+    const components = ct.components.filter(part => !this.partRule(part, 'hidden'));
```
maxbates: use block.isHidden() in case this changes

duncanmeech: fixed

src/containers/graphics/views/layout.js
duncanmeech: fixed

PR #452, opened 2016-05-20
commit name/description properties when unfocusing inputs in inspector

duncanmeech:


maxbates:
will review actually - input should be doing this

maxbates:
not necessary any more after #egfad-484-360

PR #455, opened 2016-05-20
limit length of text inputs for various important places

duncanmeech:
Block names: 64 chars
Block Description: 1024 chars
Project Title: 256 chars
Project Description: 2048 chars
Search Box: 128 chars


maxbates:
these should go in the schemas. i can do that.

PR #459, opened 2016-05-23
Inspector tweaks

maxbates:


src/styles/ListOption.css
```diff
@@ -9,6 +9,7 @@
   cursor: pointer;
   padding: 2px 5px;
   margin: 1px 0;
+  user-select: none;
```
duncanmeech: Not in love with excessive use of this, what if I want to cut and paste the name of some block into google etc?

PR #460, opened 2016-05-23
persist inspector/inventory/inventory tab state and reload on app start

duncanmeech:


maxbates:
gosh, have you even heard of reducer enhancers?

PR #461, opened 2016-05-23
autofocus search, at least when inventory already open

maxbates:


duncanmeech:
amazingly this two liner has a merge conflict

PR #463, opened 2016-05-24
make dropping of sbol block identical to regular blocks

duncanmeech:
Identical to regular blocks now. If the target block has sequence it will get pushed down along with the new block. If the block is empty ( e.g. [New Construct] ) then it will just get added to the end of the blocks current components.

### DIFF
--- a/src/containers/graphics/views/constructviewer.js
+++ b/src/containers/graphics/views/constructviewer.js
@@ -13,7 +13,6 @@ import {
   blockAddComponents,
   blockClone,
   blockSetRole,
-  blockAddSbol,
   blockRename,
   blockRemoveComponent,
 } from '../../../actions/blocks';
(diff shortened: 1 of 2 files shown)


src/containers/graphics/views/constructviewer.js
```diff
@@ -404,27 +404,38 @@ export class ConstructViewer extends Component {
...
+            this.props.blockAddComponent(oldParent.id, block.id, replaceIndex);
+            // now add the two blocks to the new parent
+            this.props.blockAddComponents(block.id, [parent.id, droppedBlock.id]);
+          } else {
+            // we can just add the dropped item into the components of the parent
+            this.props.blockAddComponent(parent.id, droppedBlock.id, parent.components.length);
+          }
         }
-        // drop on existing block
-        const newBlock = this.props.blockAddSbol(insertionPoint.block, item.id);
```
maxbates: This action can be deleted too right?

duncanmeech: done

PR #486, opened 2016-06-03
Instance Caching

maxbates:
Sets up infrastructure for blocks and projects to be stored outside of the store. 

instanceMap holds the latest version of the block / project, and is used for caching when saving projects. 

For now, you should still use the store, because if the user does anything with Undo, the store is updated, but not instanceMap.

instanceMap is useful for things like inventory - which defualts to the store, but if the block isn't there, then it looks in this map. That way, blocks dont need to be added to the store, but can be added to the map, so the store can be cleaned.


src/actions/projects.js
```diff
@@ -131,12 +131,39 @@ export const projectLoad = (projectId) => {
...
+    }
+
+    return dispatch(projectSave(currentProjectId))
       .then(() => {
-        //dont need to load the project, projectPage will handle that
-        const projectId = !!inputProjectId ? inputProjectId : getItem(recentProjectKey);
-        //alternatively, we can just call react-router's browserHistory.push() directly
+        /*
+        future - clear the store of blocks from the old project.
+        need to consider blocks in the inventory - loaded projects, search results, shown in onion etc. Probably means committing to using the instanceMap for mapping state to props in inventory.
```
maxbates: this has implications to the inspector (inspector exects blocks to be in the store), so holding off on this.

PR #487, opened 2016-06-03
Ordering Roundtrip connected to EGF

maxbates:
includes #487


server/order/egf.js
```diff
@@ -5,7 +5,7 @@ import { headersPost } from '../../src/middleware/headers';
 //testing
 import { fileWrite } from '../utils/fileSystem';
 
-const url = 'http://synnp.org:8010/';
+const url = 'http://ec2-52-30-192-126.eu-west-1.compute.amazonaws.com:8010/api/order/';
```
duncanmeech: ENV_VAR

server/order/index.js
```diff
@@ -64,22 +64,21 @@ router.route('/:projectId/:orderId?')
               projectVersion: sha,
               status: {
                 foundry,
+                response,
                 remoteId: response.jobId,
                 price: response.cost,
-                timeSent: time,
+                timeSent: new Date(time).toJSON(),
```
duncanmeech: UTC? moment.js is more reliable than Date

server/order/egf.js
maxbates: No, this is for EGF. each foundry will get its own file.

PR #493, opened 2016-06-06
Fixed Genbank I/O. All unit tests but one pass now.

florencio-mazzoldi:
There were issues with rules.role. All fixed. Tests log:
  320 passing (14s)
  22 pending
  1 failing

  1) Store Autosaving marks dirty when in between throttles:
     Uncaught AssertionError: not dirty after first save
      at [object Object]._onTimeout (autosaving.js:95:9)


florencio-mazzoldi:
@maxbates

PR #497, opened 2016-06-06
Egfad 584

duncanmeech:


src/containers/graphics/views/layout.js
```diff
@@ -135,35 +137,84 @@ export default class Layout {
...
-        text: listBlock.metadata.name,
-        fill: this.fillColor(block.id),
-        color: this.fontColor(block.id),
-        updateReference: this.updateReference,
+    // get the focused list for this block, if any
+    const focusedOptionId = this.focusedOptions[block.id];
+    // get only the options that are enabled for this block
+    const enabled = Object.keys(block.options).filter(opt => block.options[opt]);
+    // if this is a template construct and the options are empty add a single empty options block
+    if (enabled.length === 0 && this.construct.isTemplate()) {
```
maxbates: should check direct parent

maxbates:
note that not all components of templates will be list blocks. Note also that templates may have heirarchy. Seems like you've hard-coded some assumptions otherwise.

PR #498, opened 2016-06-06
Persist project version

maxbates:
project.version is written after the commit
userId used as author when init + save project

several fixes + cleanups to server filesytem utils etc.


server/data/index.js
```diff
@@ -291,6 +289,8 @@ router.route('/:projectId')
   .get((req, res, next) => {
     const { projectId } = req;
     //const { depth } = req.query; //future
+    
+    console.log('getting project ', projectId);
```
duncanmeech: my console doth overfloweth with your junk

PR #504, opened 2016-06-08
Order submit

duncanmeech:
Functionality as reviewed this morning + order DNA button on templates.


maxbates:
can you please update your babel? You are missing semicolons on class properties and it breaks the compiler. I have 6.4.5

maxbates:
uiShowOrderForm should also accept a page number, OR, if its submitted, default to page 3

src/containers/orders/selector.js
```diff
@@ -16,8 +17,9 @@ export default class Selector extends Component {
           this.props.onChange(evt.target.value);
         }}
         defaultValue={this.props.value}
+        disabled={this.props.disabled}
       >
-        {this.props.options.map(option => {
+        {this.props.disabled ? null : this.props.options.map(option => {
```
maxbates: should support a default value / placeholder text

PR #508, opened 2016-06-08
Orders inspector

maxbates:
puts orders in the inspector for projects
inspector components delegate much better, much cleaned up

### DIFF
--- a/src/styles/OrderList.css
+++ b/src/styles/OrderList.css
@@ -1,0 +1,23 @@
+@import './variables.css';
+
+.OrderList {
+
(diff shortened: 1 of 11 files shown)


src/styles/OrderList.css
```diff
@@ -0,0 +1,23 @@
+@import './variables.css';
+
+.OrderList {
+
+  &-group {
+    padding-bottom: 0.5em;
```
duncanmeech: prefer REM over EM for padding/margins but either way is good.

PR #512, opened 2016-06-09
selected option shows up in inspector

maxbates:


duncanmeech:
Shift clicking the options does not implement the range selection behavior that shift clicking the top level blocks do i.e. shift click the 1st option in the first block and the 1st option in the last block should still range select the blocks between them.

maxbates:
I think we're ok to merge this as is.... If you want to add this functionality, you can use the `forceBlocks` part of the focus - they won't be editable, but at present the idea is you can't edit option blocks anyway. But options rely on the optionId map currently to show them

PR #514, opened 2016-06-11
Image compare tests

duncanmeech:
All e2e tests refactored for all the app changes that have occurred since the last time they worked!

Also, this PR includes new image based tests. How to use/update:
**npm run nightwatch** 
This runs all tests and in some cases generated saved screenshots of the app at critical points. If all the test pass then...
**npm run imagediff**
Will run the image comparison tests against cannonical images that are part of the repository.
If all images are perfect matches the app will exit. If there are differences the app will open a browser window that displays the cannonical image and the newly generated image. You can select either 1 or 2 to pass the test and update the image or fail the test. 
NOTE: If you update the image you will need to commit/push your changes to share the new cannonical images.


maxbates:
+1 from me

PR #515, opened 2016-06-12
Egfad 634

duncanmeech:
List blocks identified with isList()...however, there don't appear to be any non list blocks in the current templates.


maxbates:
good after remove debugger statement

PR #557, opened 2016-06-17
do not allow blocks with child components to have a sequence set

duncanmeech:


src/components/importdna/importdnaform.js
```diff
@@ -42,9 +42,16 @@ class DNAImportForm extends Component {
         this.props.uiSetGrunt(`Sequence data must be added to a selected block. Please select a block and try again.`);
         return;
       }
-      if (nextProps.currentConstruct.isFrozen() || nextProps.currentConstruct.isFixed()) {
+      if (nextProps.currentConstruct.isFrozen() || nextProps.currentConstruct.isFixed() || nextProps.currentConstruct.isTemplate()) {
```
maxbates: you should be checking for if its a construct, not just a template

use isConstruct()

maxbates:
see comment

PR #565, opened 2016-06-20
Saving of blocks, format of project rollups

maxbates:
Big one.

Blocks are now all saved in a single file, so writing / loading a project is much faster

Block rollups are now objects, in the form { blockId: block } . previously they were arrays, which required more lookups and was particularly unsuitable for storing all blocks in a single file + handling merging with that file.

unit tests all pass. nightwatch still not working on my machine (not sure why)


duncanmeech:
rebase required

duncanmeech:
Merged with master but double check index.js

PR #572, opened 2016-06-20
Egfad 690

duncanmeech:


duncanmeech:
Accidentally pushed in card egfad-724 in here as well...clicking construct name.

PR #586, opened 2016-06-21
User setup - empty project

maxbates:
@duncanmeech take note how projects are added to a user's account, since you will likely add the one that eli makes if he does end up making one. 

Pretty straightforward. But do make sure you use `Block.classless()` instead of `new Block()` to avoid some issues with writing to file system (e.g. see egf starter project in /data/egf_parts/project.js)


maxbates:
hold up a second on this one

maxbates:
see #590

PR #588, opened 2016-06-21
Exports the sample project now.

florencio-mazzoldi:
There were a few different issues with the sample file.
Color should be maintained for annotations now.

Order Test Fails!
TEST FAILURE: 1 assertions failed, 520 passed (11m 23s)
 ✖ template-order
   - Verify we can order all templates
     Testing if element .order-button occurs on the page 29 times. - Expected "29" but got: "0"

### DIFF
--- a/plugins/convert/genbank/genbank_export.py
+++ b/plugins/convert/genbank/genbank_export.py
@@ -136,7 +136,7 @@ def project_to_genbank(filename, project, allblocks):
         if "genbank" in block["metadata"] and "id" in block["metadata"]["genbank"]:
             genbank_id = block["metadata"]["genbank"]["id"]
         else:
-            genbank_id = block["metadata"]["name"].replace(" ", "")
+            genbank_id = "GC_DNA"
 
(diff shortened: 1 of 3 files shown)


plugins/convert/genbank/genbank_export.py
```diff
@@ -177,7 +177,7 @@ def project_to_genbank(filename, project, allblocks):
         if "description" in block["metadata"]:
             seq_obj.description = block["metadata"]["description"]
         if "name" in block["metadata"]:
-            seq_obj.name = block["metadata"]["name"].replace(" ", "")
+            seq_obj.name = block["metadata"]["name"].replace(" ", "")[:5]
```
maxbates: why? did you mean sequence?

florencio-mazzoldi: No, the name in biopython accepts a short amount of characters, so I just truncate the name of the construct as we export. I could have put "unknown" or something like that. We can change it if it's a problem.

PR #591, opened 2016-06-21
Projects querying with grep doesn't error

maxbates:
querying user projects uses spawn rather than exec, does not error when user has no projects. add a test

includes part of #590 (merge that first)


duncanmeech:
Looks good, rebase then merge

PR #592, opened 2016-06-22
generate and require a label for each order.

duncanmeech:


src/containers/graphics/views/constructviewer.js
```diff
@@ -516,8 +520,12 @@ export class ConstructViewer extends Component {
    * launch DNA form for this construct
    */
   onOrderDNA = () => {
-    const order = this.props.orderCreate(this.props.projectId, [this.props.construct.id]);
-    this.props.uiShowOrderForm(true, order.id);
+    let order = this.props.orderCreate(this.props.currentProjectId, [this.props.construct.id]);
+    this.props.orderList(this.props.currentProjectId)
```
maxbates: why?

PR #593, opened 2016-06-22
disable copy on fixed / frozen constructs ( so they can not be pasted…

duncanmeech:
… later )


maxbates:
copying should be ok, its pasting that you want to block. when you clone a block, it is no longer frozen.

PR #604, opened 2016-06-23
various small fixes + tweaks

maxbates:


src/containers/orders/checkbox.js
```diff
@@ -5,7 +5,8 @@ import '../../../src/styles/ordermodal.css';
 
 export default class Input extends Component {
   static propTypes = {
-    value: PropTypes.string.isRequired,
+    value: PropTypes.bool.isRequired,
```
maxbates: note that i changed this to a boolean. it also returns a boolean on changes. this makes more sense.than strings.

PR #620, opened 2016-06-30
selecting an option does not select the block

duncanmeech:
( although you can still select a block by dragging a rectangle around the options which is nice ).


src/containers/graphics/views/constructvieweruserinterface.js
```diff
@@ -391,7 +391,6 @@ export default class ConstructViewerUserInterface extends UserInterface {
       case 'option':
         // user might have clicked an 'empty list' placeholder, otherwise select the option
         if (region.optionId) {
-          this.constructViewer.blockSelected([block]);
```
duncanmeech: This is basically the click handler for the block. If they clicked in an option ( case 'option' ) then the option is selected but not the block, so top level block selections are unchanged.
If you click the top of the block it selects as normal ( and unselects other blocks unless you are shift clicking ).

PR #636, opened 2016-07-12
Egfad 793

duncanmeech:


maxbates:
process.env.NODE_ENV !== 'production'

duncanmeech:
Comment addressed

server/server.js
```diff
@@ -135,7 +136,7 @@ app.get('*', (req, res) => {
       discourseDomain: process.env.BNR_ENV_URL_SUFFIX || `https://forum.bionano.autodesk.com`,
     };
     //so that any routing is delegated to the client
-    res.render(path.join(pathContent + '/index.jade'), Object.assign({}, req.user, discourse));
+    res.render(path.join(pathContent + '/index.jade'), Object.assign({}, req.user, discourse, {production}));
```
maxbates: Why do you need to re-expose this? everything else just uses `process.env.NODE_ENV === 'production'` and the like

PR #642, opened 2016-07-14
Several annotation enhancements.

florencio-mazzoldi:
All genbank imports start from the same color now.
Annotations have roles.
Annotations have colors (automatically assigned), and they roundtrip them through Genbank.
Flattened out the color attribute for the annotation. It's on the main structure, not in the metadata.


maxbates:
@florencio-mazzoldi can you verify your tests / look this over?

PR #643, opened 2016-07-14
add empty block as an available sbol type, remove the old no symbol o…

duncanmeech:
…ption from the picker ( projects with that type will still load ). Update tests and image tests


src/inventory/roles.js
```diff
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 export const symbolMap = {
+  emptyBlock: 'Empty Block',
```
maxbates: This should be No Symbol

maxbates:
No Symbol =/=> Empty. They can still have a sequence.They just dont have a role. Rename accordingly

This should never set block.role when you select the no symbol. You can set block.rules.role to null if you want to set it to something. Ideally, delete block.rules.role

maxbates:
Also, not sure why you removed the no symbol SVG. I think you want to just use that one. I think Joe was confused when he specified it as blank, as we added the no symbol glyph for that very reason

PR #647, opened 2016-07-15
give roles their own inspector + update focus to handle them

maxbates:
Currently can't inspect roles. fixes that.


duncanmeech:
user-select: none 
on the contents makes for a bad experience. Users should be able to select and copy text from the inspectors.

PR #674, opened 2016-08-01
add doc on ssh into instance

maxbates:


florencio-mazzoldi:
Not sure it's ok to put our IP addresses there..

maxbates:
Drew said it was ok, since you have to be on the network to access anyway

PR #684, opened 2016-08-05
[tag] 0.6.0 initial extensions framework

maxbates:
Beginning better versioning. Marking this point as largely stable, with initial version of extension framework


PeterRJones:
Looks good @maxbates build and deploy pipeline should pick it up with no issues.

PR #686, opened 2016-08-06
Seq viewer test

duncanmeech:
#### Overview

Adds a basic fabric test for the new sequence viewer

#### Type of change (bug fix, feature, docs, etc.)

e2e test ( also improvement to reset/forgot password test )

#### Requirements


maxbates:
I think we need a better solution for testing extensions. I dont think the tests belong in the main repo

maxbates:
You can't test that it will be in the app, because there will be scenarios when it is not. E.g. user-preferences that specify not including it.

duncanmeech:
..but we don't have that option right now. So testing in situ is appropriate.

maxbates:
once user prefs are in (in the next week or so) you will have to fix this to check for it being present

PR #694, opened 2016-08-09
menus dont dissappear automatically, they have a timeout

maxbates:
#### Overview

Menus timeout and disappear on mouse-leave after 300ms

#### Type of change (bug fix, feature, docs, etc.)

Feature

#### Issue

EGFAD-923


duncanmeech:
Did you run e2e on this? Seem possibly it might break some of them. Do you test with context menus etc? Just saying...

PR #710, opened 2016-08-12
fix reset password page

duncanmeech:
#### Overview

API was moved from data to auth. Code was not updated. It is now.

#### Requirements


maxbates:
test is ok

PR #713, opened 2016-08-15
Add GSL

maxbates:
#### Overview

add gsl extension

Featuresss


maxbates:
they are in the dockerfile

PR #715, opened 2016-08-15
Egfad 936

duncanmeech:
#### Overview

Fix most if the outstanding issues with sequence editing.

- Honor the case of the sequence in the sequence viewer AND the sequence editor.
- Only allow single, user defined, blocks to be edited.
- Display the current sequence for blocks that are being edited.
- Remove the old functionality for pushing down parents, secretly creating unexpected new blocks etc.

EGFAD-936
EGFAD-914

#### Requirements


src/components/importdna/importdnaform.js
```diff
@@ -68,11 +70,21 @@ class DNAImportForm extends Component {
...
+      }
+      block.getSequence()
+        .then(sequence => {
+          this.setState({
+            inputValid: true,
+            validLength: sequence ? sequence.length : 0,
+            sequence: sequence || '',
+            haveSequence: true,
+          });
+        });
```
maxbates: you should add a catch to make sure always will show modal
not sure when you would get an error

Issue #79, opened 2015-11-18
update redux-devtools (pending beta completion)


maxbates:
https://github.com/gaearon/redux-devtools/releases
https://github.com/gaearon/redux-devtools/pull/132

see for implementation: https://github.com/erikras/react-redux-universal-hot-example/

PR #722, opened 2016-08-16
better check port on server start

maxbates:
#### Overview

when checking for open port, check against hostname instead of localhost

Bug fix


dhylbert:
I feel like this code changes doesn't really match the description.

maxbates:
port = host + port :)

PeterRJones:
If 'hostname' is used in listen({ port, host: hostname, exclusive: true,}) then line #79 should be updated to match. Confirm that line 179 stays in sync as well.

PR #726, opened 2016-08-16
enable docker container restart in dev/qa/prod

dhylbert:
#### Overview

Tell docker-compose to restart the gctor docker container when it exits no matter the reason. Even though the `production` compose file inherits from the `quickstart` compose file, I've added it to both places in case we want to change the functionality in dev/qa, and to increase verbosity.

runtime configuration change

#### Breaking Changes

None

#### Requirements


dhylbert:
This change should be merged into `gsl`.

PR #732, opened 2016-08-17
Inventory item load

maxbates:
#### Overview

UX updates for inventory items when loading, loading fails

Feature

#### Other information

Handles when fetch from NCBI doesnt work, and when conversion fails. makes drag-drop not active. doesnt allow converting already converted inventory items


src/components/Inventory/InventoryItem.js
```diff
@@ -154,14 +200,16 @@ export class InventoryItem extends Component {
     return (
       <div className={'InventoryItem' +
       ((!!image || !!svg) ? ' hasImage' : '') +
+      ((!!loading && !skipFocus) ? ' loading' : '') +
+      ((!!loadError) ? ' loadError' : '') +
```
duncanmeech: I have a team of six guys trying to figure out what this does....I'll let you know what they find.

PR #740, opened 2016-08-24
hidden block handling

duncanmeech:
#### Overview

Correctly handle hidden blocks in nested constructs including in authoring mode where hidden blocks are always displayed.

New Display Rules:

1. A hidden parent ( and its children ) is only displayed in authoring mode.
2. A parent with all hidden children will not display any children ( or a connector )
3. A parent with some hidden children will attach the connector to the first visible child ( which varies with authoring mode ).

#### Requirements


src/containers/graphics/views/constructviewer.js
```diff
@@ -519,6 +520,8 @@ export class ConstructViewer extends Component {
           const oldParent = parent;
           parent = this.props.blocks[insertionPoint.block];
           if (parent.hasSequence()) {
+            // ensure the parent isn't hidden ( sub trees containing a mix or hidden / unhidden blocks cause problems )
+            this.props.blockSetHidden(parent.id, false);
```
maxbates: You're changing the parent? In any case when dropping onto a hidden block? I don't think this is ok. You need to be able to deal with children being a mix of hidden and not hidden...

duncanmeech: Good catch, that was my first attempt at fixing the problem. Forgot to remove it.

PR #741, opened 2016-08-24
Egfad 969

duncanmeech:
#### Overview

Allow moving and/or clone + move of blocks when in authoring mode, regardless of isFixed().

#### Requirements


maxbates:
seems duplicate of #740

PR #753, opened 2016-09-07
Gb name

florencio-mazzoldi:
#### Overview
Several enhancements for import and export of Genbank.
  - Name is imported based on the genbank standard
  - "label" is still supported because it's what Ape uses
  - Annotation start and end are now keyed to the start and end of the block they belong to
  - All our data is stored in the "note" feature of the exported genbank file
Added an e2e test for ncbi import.

Fix

#### Breaking Changes
None - All import/export tests pass.

#### Requirements

- [X] Adds a test (for features + fixes)


maxbates:
failing test is ok (not relevant)

PR #948, opened 2016-09-20
Zube 905 Correctly set sequence edit box after sequence loaded.

duncanmeech:
#### Requirements


maxbates:
see #953

PR #949, opened 2016-09-20
Zube 945 Can edit sequence for non list blocks in authoring templates. ( but not egf )

duncanmeech:
#### Requirements


maxbates:
see #953

PR #953, opened 2016-09-20
Fix lint ( based of todays PR's )

duncanmeech:
#### Overview
This is all of todays PR's + lint fixes. If you approve them then one is the only one that needs to be merged.

#### Requirements


test-e2e/nightwatch.js
```diff
@@ -10,18 +10,18 @@ module.exports = {
...
   selenium: {
     start_process: true,
-    server_path: './node_modules/selenium-standalone/.selenium/selenium-server/2.53.1-server.jar',
+    server_path: './node_modules/selenium-standalone/.selenium/selenium-server/2.50.1-server.jar',
     log_path: '',
     host: '127.0.0.1',
     port: 4444,
     cli_args: {
-      'webdriver.chrome.driver': './node_modules/selenium-standalone/.selenium/chromedriver/2.24-x64-chromedriver',
+      'webdriver.chrome.driver': './node_modules/selenium-standalone/.selenium/chromedriver/2.21-x64-chromedriver',
```
maxbates: see #932 , these are the current versions and are pinned properly in package.json on that PR. just need to merge properly, and that PR is correct

maxbates:
All the urls at `:3000` work for the local tests when browsersync is not available and a simple instance is started. We can do an env var, as you can already pass in `PORT=3000` for the server, so just pick that up in your tests as well?

maxbates:
made tests run at port 3001

PR #983, opened 2016-09-29
Roundtrip keeps the hierarchy

florencio-mazzoldi:
#### Overview
Storing block id and parents in the genbank file and using them to re-create the hierarchy.
Also fixed a bug when storing annotations.

Both feature and bug fix.

#### Breaking Changes
None. Ran all tests and look good. There are 2 tests for the overall API that fail, @maxbates 

#### Requirements


server/extensions/native/genbank/genbank_import.py
```diff
@@ -271,6 +278,14 @@ def has_parent(block_id, all_blocks):
             return True
     return False
 
+def block_by_old_id(old_id, all_blocks):
+    for block in all_blocks.values():
+        if "old_id" in block["metadata"] and block["metadata"]["old_id"] == old_id:
+            print "returning", block
```
cojofra: print

PR #984, opened 2016-09-29
Inventory defaults to open for new users

maxbates:

feature


src/reducers/ui/inventory.js
```diff
@@ -17,7 +17,7 @@ import * as ActionTypes from '../../constants/ActionTypes';
 import { getLocal, setLocal } from '../../utils/ui/localstorage';
 
 export const initialState = {
-  isVisible: getLocal('inventoryVisibility'),
+  isVisible: getLocal('inventoryVisibility') || true,
```
duncanmeech: getLocal accepts the default return value...so you should write...
getLocal('inventoryVisibility', true);

PR #989, opened 2016-09-29
new blocks/constructs always show as new block

duncanmeech:
#### Overview

- unnamed blocks now appear as New Block in construct viewer and inspector.
- blocks and construct just appear as Block / New Block in the inspector ( except for list blocks and fillers ).

#### Requirements


src/containers/graphics/views/layout.js
```diff
@@ -450,7 +450,7 @@ export default class Layout {
    * If the part is an SBOL symbol then use the symbol name preferentially
    */
   partName(part) {
-    return this.blocks[part].getName('Block', true);
+    return this.blocks[part].getName('New Block', true);
```
maxbates: this shouldn't be necessary

PR #991, opened 2016-09-29
Python extension

maxbates:
#### Overview

example extensions which calls an external script

example


src/reducers/ui/inventory.js
```diff
@@ -17,7 +17,7 @@ import * as ActionTypes from '../../constants/ActionTypes';
 import { getLocal, setLocal } from '../../utils/ui/localstorage';
 
 export const initialState = {
-  isVisible: getLocal('inventoryVisibility'),
+  isVisible: getLocal('inventoryVisibility', true),
```
duncanmeech: I already fixed that and the line below. Will probably merge fine

PR #992, opened 2016-09-29
sequence routes can accept database URLs outside production

maxbates:
#### Requirements


maxbates:
after initial testing, this is much slower

PR #993, opened 2016-09-30
removing prior list options from focus-block-options action

duncanmeech:
#### Overview

Removing old list options from FOCUS_BLOCK_OPTION action. They just accumulate over time and confused listeners to the action.

#### Requirements


maxbates:
this is incorrect, and will reset all other options when you make a change.

maxbates:
alternatively, move this logic to the reducer.

PR #994, opened 2016-09-30
render 404 as a view within the app rather than use an iframe from th…

duncanmeech:
#### Requirements

#### Issue

…e server

### DIFF
--- a/src/styles/404.css
+++ b/src/styles/404.css
@@ -1,26 +1,4 @@
-
 @import './variables.css';
-
-
-@font-face {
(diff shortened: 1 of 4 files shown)


src/styles/404.css
```diff
@@ -1,5 +1,27 @@
+
 @import './variables.css';
 
+
+@font-face {
```
maxbates: this should go in a fonts.css file, and be included in the main css module

duncanmeech: done

PR #1052, opened 2016-10-17
Use s3 - sequences, jobs, project files

### LINKED ISSUE #870: (test) project files

### LINKED ISSUE #913: Import middleware - batch sequence writing
to ensure that we dont have too many files open, on very large imports.

Once done, either genbank should delegate sequence writing to this, or make new issue to do that.

### LINKED ISSUE #1027: Genbank converter should return sequences on rollup
Genbank conversions writes sequences itself. It should return them on the rollup.

This may have been done somewhere else? or already in?

### LINKED ISSUE #1041: Genbank uses sequence range syntax
Migrate the genbank conversions so they write a single big file, rather than many files, and use the format added by #1040

python code breaks up the sequences, and instead should return a large sequence, and the appropriate ranges for each block

### LINKED ISSUE #1044: project files should be tracked in the project JSON
Project files need to be tracked in the project JSON and, when written (new file each time) should save the project as well

### LINKED ISSUE #1045: Sequences written to S3 in production
Following #1034, which add strucutre for using s3 in production and file system locally, use S3 for file writes
- [x] sequenceWrite
- [x] sequenceWriteMany (how to handle batching?)
- [x] sequenceRead
- [x] sequenceRead byte range
- [x] sequenceExists (probably want to remove use in code where possible)

### LINKED ISSUE #1046: Sequences can be written / accessed by the client
After #1045, instead of going through the application server, the client should be able to get an access token to read and write sequnces on its own

### LINKED ISSUE #1051: Schema validation - Block md5 format
Rather than just string validation, should check that the format is valid (`md5[start;end]` - see regex in `src/utils/sequenceMd5.js`)

### LINKED ISSUE #1055: S3 tests all are prefixed into a test folder
Can set prefix in test environment so easy to clear out test files

maxbates:
#### Overview
- job files are not really handled... tmp files used still - need to clarify where these are going
- [ ] clarify use for job files
- [x] Sequences are written in S3 in production (or when creds provided)
- [x] project files saved in s3 (when creds provided)
- [x] project files are saved on the project
  - [x] actions, which update the project
  - [x] router, which writes input + output S3
  - [x] update `constructor.extensions.files`
  - [x] tests (actions)
- [x] update travis file to matrix tests, one set using S3 and one set not. several unit tests.
- [x] move sequence operations outside of persistence master export. 
  - [x] In tests
  - [x] in routers etc.
- ~~mirror job files in S3 (later, will have client handle upload, and jobs will download -- for now just go through the server)~~ Job files are just sent in POST message bodies?
- [x] prefix S3 buckets with `test` 
- [x] verify S3 buckets set up properly when testing
- ~~clear s3 test buckets after tests~~ see #1111 
- [x] verify HOST_URL in project file router as correct in production
- [x] Genbank conversions use the new format ( #1041 1041)
- [ ] migrate existing sequences / project files ???

Follows #1040

feature
#### Breaking Changes
- expects `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` in production. 
- projectFilesWrite() return { url, VersionId }
- projectFilesList() returns [ {name, url } ]
#### Requirements
#### Other information
- s3 configuration: can pass `FORCE_LOCAL` to avoid using S3 even if credentials passed. Can pass `AWS_S3_LOCATION` or defaults to `us-west-1`
#### Issue

close #870 
close #913 
close #1027 
close #1041 
close #1044
close #1045 
close #1046 (kinda)
close #1055 
close #1051


maxbates:
The only test failing at this point is a large genbank conversion - it fails because writing that many files takes too long  with S3

will need to tackle #1041 simultaneously, so only one file written at a time

server/data/persistence/s3.js
```diff
@@ -0,0 +1,138 @@
...
+if (process.env.NODE_ENV === 'production' || (
+    (!process.env.FORCE_LOCAL || (!!process.env.FORCE_LOCAL && process.env.FORCE_LOCAL !== 'true')) &&
+    (process.env.AWS_ACCESS_KEY_ID && process.env.AWS_SECRET_ACCESS_KEY)
+  )) {
+  invariant(!!process.env.AWS_ACCESS_KEY_ID, 'expected env var AWS_ACCESS_KEY_ID');
+  invariant(!!process.env.AWS_SECRET_ACCESS_KEY, 'expected env var AWS_SECRET_ACCESS_KEY');
+
+  AWS = require('aws-sdk');
+
+  AWS.config.update({
```
duncanmeech: You can get the region from the API, better than hard coding since you don't know which region the credentials refer to.

maxbates: how do i do this?

maxbates:
unit tests run in both S3 and against file system in Travis

server/data/persistence/s3.js
duncanmeech: good question. For buckets you can use: http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#getBucketLocation-property
But if we are going to create a bucket you will need to specify a region, so probably an ENV_VAR would be necessary.

maxbates:
This branch now includes several changes for project-files in s3, and more tests against s3

PR #1076, opened 2016-10-21
removed sudo references in Dockerfile

PeterRJones:
#### Overview
The sudo command is typically not present within the context of a Dockerfile and thus, when referenced, break the docker build.

Dockerfile commands run as root (unless a USER directive has been made previously in the Dockerfile). Thus commands that require root level privilege, such as apt-get, do not need to have `sudo` in front of these commands.

Note that this is different than when these commands run interactively within a Linux shell, **unless you are running as root** within the Linux shell. 

You can see whom are you running as in Linux via the `whoami` command.
You can run interactively as root via the `sudo -i` command. When finished running root-level commands, use `exit` to stop the root-level shell and return to your normal interactive user shell. 

bug fix

#### Breaking Changes
None

#### Requirements

#### Other information

See description above.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,8 +15,8 @@ RUN apt-get update -y
 RUN apt-get install -y curl gcc-4.9 g++-4.9 libstdc++-4.9-dev
 
 RUN apt-get install -y python python-dev python-pip git build-essential wget && \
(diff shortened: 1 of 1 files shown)


PeterRJones:
FYI @maxbates @duncanmeech

PR #1091, opened 2016-10-25
Recommend devs use `npm run start-instance`

BevanR:
#### Overview

Recommend devs use `npm run start-instance` instead of `npm run start` (`npm start`).

`npm run start` doesn't start a webserver.  `run-instance` does.

Documentation only.

### DIFF
--- a/docs/installation/installation.md
+++ b/docs/installation/installation.md
@@ -20,7 +20,7 @@ Easiest way to get set up, using mocked authentication. In this installation, th
 
 `npm install`
 
-`npm run start`
+`npm run start-instance`
 
 ### In a Docker Container
 


maxbates:
Hi! `npm run start` should be starting a web server, it takes about 30 seconds. `npm run start` is recommended because it creates a server which will automatically reload when file changes are made, where as `npm run start-instance` will not.

If you are not getting a server running, can you post what output appears in your terminal?

BevanR:
You are correct.  The first time it didn't start a server though.  I still have it in my bash session to prove it!  :)

screen shot 2016-10-25 at 3 00 42 pm

Is it possible that running `npm test` in another bash session conflicted with `npm start`?

maxbates:
Yes - the server checks if the port is free, and if not... apparently it dies silently and languishes (server/server.js L175 or so)

BevanR:
Ah!  That explains it.

BevanR:
Thank you!  :)

PR #1227, opened 2016-12-15
models and schemas are bundled into a package

### LINKED ISSUE #1015: Break out Models into npm module
Particularly in extensions, access to models is important. Can't refer to file-system, so may make sense to break out the models into their own module of the app explicitly, and host an NPM package for them

maxbates:
#### Overview

Puts Block, Project, Order, Rollup models and schemas into a separate bundle. Can publish this bundle on npm so it can be consumed separately, e.g. by extensions

This is just the setup + scripts. it is not actually pushed yet.

feature

#### Breaking Changes

none

#### Requirements

#### Issue

close #1015


maxbates:
merging into work on blast

PR #1232, opened 2016-12-16
simplify local authentication start script

dhylbert:
The move to DB-backed project storage changes the way a user would enable DB-backed authentication for local development. The `npm run auth` target assumes a user would have the entire `bio-user-platform` application running, including DB, and then the would need to override the DB password used by GCTOR storage to save projects.

To make this more simple, I removed the `npm run auth` target; I'm probably the only person that would have a `bio-user-platform` stack running locally. I then updated the `npm run start-auth` procedure to install the authentication module from the directory to which the `bio-user-platform` user platform is cloned locally. I also fixed some log messages to make them more clear.

### DIFF
--- a/tools/auth-stack.js
+++ b/tools/auth-stack.js
@@ -32,14 +32,14 @@ const setupBioNanoPlatform = (useGenomeDesignerBranch = false) => {
   const checkoutPromise = useGenomeDesignerBranch === true ?
     promisedExec(`git checkout genome-designer`,
       { cwd: pathBioNanoPlatform },
-      { comment: 'Setting up User Module...' }
+      { comment: 'Checking out \'genome-designer\' branch of User Platform...' }
     ) :
     Promise.resolve();
 
   return checkoutPromise
     .then(() => promisedExec(`npm install`,
       { cwd: pathBioNanoPlatform },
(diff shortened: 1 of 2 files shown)


tools/auth-stack.js
```diff
@@ -50,15 +50,28 @@ const startAuthServer = () => {
...
+
 const startRunAuth = () => {
   console.log('\n\n');
-  return spawnAsync('npm', ['run', 'auth'],
-    { cwd: pathProjectRoot },
+  return spawnAsync('npm', ['run', 'start'],
+    { cwd: pathProjectRoot,
+      env: Object.assign({
+        BIO_NANO_AUTH: 1,
+        HOST_URL: 'http://localhost:3000',
```
maxbates: I'd recommend you use the HOST_URL constant? /server/urlConstants.js

tools/auth-stack.js
```diff
@@ -50,15 +50,28 @@ const startAuthServer = () => {
...
+      }, process.env),
+    },
     {
       comment: 'Starting Constructor with Authentication...',
       waitUntil: 'Server listening at http://0.0.0.0:3000/',
```
maxbates: this too should probably by dynamic...

PR #1277, opened 2017-01-04
Various fixes for the sketch panel including removing filtering and changing the order of the blocks

duncanmeech:
#### Overview

https://zube.io/autodesk/genome-designer/c/1249

#### Requirements


src/components/Inventory/InventoryGroupRole.js
```diff
@@ -115,21 +106,49 @@ export default class InventoryGroupRole extends Component {
     return proxy;
   }
 
+  /**
+   * dictates the sort order of sbol tiles
+
+   */
+  static sortOrder = [
```
maxbates: This should be part of roles.js

PR #1281, opened 2017-01-04
Sequence viewer options block select fix

duncanmeech:
#### Overview

https://zube.io/autodesk/genome-designer/c/1137

Let extensions know which was the last selection option versus just sending all the currently selected options/blocks ids

#### Requirements


src/actions/focus.js
```diff
@@ -231,6 +231,10 @@ export const focusBlockOption = (blockId, optionId) => {
     const oldOptions = getState().focus.options;
     const options = Object.assign({}, oldOptions, { [blockId]: optionId });
 
+    // options is all the focused blocks and associated selected option. For extensions etc
+    // we need to show which block was most recently selected.
+    options.selectedOptionId = optionId;
```
maxbates: This is already available at focus.optionId - you can't add this to the map as it is used many places and expects only blockIds

maxbates:
not as is

maxbates:
I'm watching your `!important` in CSS

src/actions/focus.js
duncanmeech: If it exists it is not being passed to the extensions in the focus object. There is focus.options but it does not give any indication of the most recently selected.

PR #1291, opened 2017-01-06
Styleguide update

### LINKED ISSUE #1282: Lint update - react rules, es6, imports

maxbates:
#### Overview

Overhaul code base for new lint rules. Many new lint rules since added linting ~ 12 months ago, especially for react specific rules and es6 and imports

Rebases to use airbnb style guide (https://github.com/airbnb/javascript), but more forgiving

style

#### Breaking Changes

none

#### Other information

all tests pass.

Noteable changes include:

- import order (node, project, local). no duplicates.
- avoid fn.bind() in JSX. Fine elsewhere. arrow functions ok when pass parameter.
- omit unused react prop types
- prefer static methods on classes
- use string templates

Forthcoming:

- no string refs. they are soon to be deprecated. pending code mod.

#### Issue

close #1282


duncanmeech:
Personally, I'm not sure that some random dude at AirBnB is necessarily the source of canonical truth on good style. The rule about bind is especially dumb and pointless. Also you ui-stylish branch was badly broken after switching to arrow functions. You should always run the e2e tests before submitting for review. Regardless, my most recent branch contains fixes for the stuff you broke so this branch is fine with me.

PR #1310, opened 2017-01-17
New search panel

duncanmeech:
#### Overview

https://zube.io/autodesk/genome-designer/c/1176

#### Requirements


maxbates:
Why did coverage go down so much?

src/containers/graphics/views/dropTarget.js
```diff
@@ -0,0 +1,110 @@
...
+import { blockClone, blockCreate } from '../../../actions/blocks';
+import { focusConstruct } from '../../../actions/focus';
+import { projectAddConstruct } from '../../../actions/projects';
+import { block as blockDragType } from '../../../constants/DragTypes';
+import DnD from '../dnd/dnd';
+import ConstructViewer from './constructviewer';
+
+import '../../../styles/droptarget.css';
+
+export class DropTarget extends Component {
```
maxbates: can you name this InterConstructDropTarget or something

maxbates:
Please address the drop in coverage.

Re projectAddComponents() and forcing the projectId  --- it will error if there is a projectId set on the block, and it is different than the project you are dropping into. When you clone, it should unset the projectId. When you are handling drops from the inventory, you need to make sure the projectId is valid throughout (set when in project A, unset when clone, set to project B when drop on project B). To handle the situation of dragging from project A to B without cloning, I'm not sure if removing the blocks from the project will unset the projectId - I think it might just on the construct - but we need to make sure that it is in fact removing projectId from all blocks being transferred.

src/containers/graphics/views/dropTarget.js
duncanmeech: There is another PR after this ( new-new-color-picker-picker-picker ) which has this fix.

PR #1316, opened 2017-01-18
Landing Page + Modal & Form Framework 

maxbates:
#### Overview

- Landing Page ( #1301 )
  - Update routing
  - put content in frame so easy to update
  - save account type to user data (free vs paid)
- New Modal framework ( #1302 )
  - update register, signin, forgot password
  - others will be relatively straightforward
- New Form styling framework
  - several components for forms + styling

#### Type of change (bug fix, feature, docs, UI, etc.)

feature

#### Breaking Changes

Changes default passwords for tests - the modal requires alphanumeric, length > 8 (but not server API)

#### Requirements

- [x] Adds a test (for features + fixes)
- [ ] Docs updated (for features + fixes)

#### Other information

e2e tests updated

#### Issue

close #1301 
close #1302
close #1305 
close #1319

PR #1336, opened 2017-01-24
Landing page separate

### LINKED ISSUE #1321: Make Landing Page crawlable
1) (Easier?) Sniff out agent, if a robot, send the landing page static content

2) (Harder) Render react on the server and send it

### LINKED ISSUE #1322: Landing page can stand-alone
Currently, landing page is in an iFrame. We should handle the case where its not, and redirect to the real genetic constructor domain, rather than always expect it is in an iframe... in case it gets served weird or something

### LINKED ISSUE #1331: Hitting "return" after typing password in new sign in dialog doesn't submit the form

maxbates:
#### Overview

following #1316 

Makes the landing page separate from the rest of the app. No more iframe. 

Users will see the landing page if they request the site root, and they are not logged in. Closing the modals will redirect back to the landing page.

Homepage now obviously crawlable yay

Authentication forms are just forms, and are wrapped in a modal by the auth form modal wrapper

fix / feature

#### Breaking Changes

landing page now separate from the app, the landing page links to modals, but does not trigger them directly

#### Requirements

#### Issue

close #1321
close #1322 
close #1331


cojofra:
landing page design ok

PR #1392, opened 2017-02-01
Fix: setting project Id on block clone

maxbates:
#### Overview

Properly sets project ID when cut / pasting hierarchical blocks

Also refactors autosave + undo parts of the store so that autosave is not triggered in the middle of a transaction, when it cannot be as guaranteed the state is correct.

Adds auto save reducer `store.autosave = { dirty: true|false, timeUnsaved: int }`

fix

#### Breaking Changes

Removes autosaveInstance.isDirty(), moves autosave status into store

#### Requirements


maxbates:
problems are lint issues, which have been resolved in ui-e2e i believe, but are not due to this PR

PR #1610, opened 2017-02-24
Webpack svg

maxbates:
#### Overview

- Inline several SVG images in package, so not loaded on page load -- main logo + section icons in inspector + inventory
- adds package svg-url-loader to webpack
- requires SVGs are used as backgroung images
- inline toolbars are pending #1611 

perf

#### Requirements

#### Issue

address #1602


duncanmeech:
Code is fine but reassigning to @jlachoff since you added an transition animation and joe will need to approve that.

jlachoff:
@duncanmeech can you please point me to a specific example on a specific build where I can review this unapproved-by-Joe transition animation? thanks!

PR #1702, opened 2017-03-21
move cookie banner down to not block sign-in

cojofra:
#### Overview
^

UI bug fix

#### Requirements


cojofra:
@jlachoff FYI

maxbates:
this will continue to hover at 69px from top after you scroll....

cojofra:
yeah but that's fine cause it just sits under the menu bar which is also scrolling along

PR #1707, opened 2017-03-22
Changes to docker and jenkins-docker files to support deployment into ecs

PeterRJones:
#### Overview
This change updates docker-compose and jenkins-docker config files so that gctor can be deployed into an AWS ECS cluster. 

deployment change

#### Breaking Changes
None

#### Requirements
None

#### Other information
This change is backward compatible with deployment gctor to existing aws/docker environments.

#### Issue

1. We now must declare maximum run-time memory needed by gctor. This is currently set to 8GB, which is too high. This should be optimized and this setting changed to reflect actual need.
2. Application startup time is > 2 minutes, which can delay registration of new tasks during autoscaling. This startup time should be optimized to be < 90 s.


PeterRJones:
//cc @dhylbert Please review

dhylbert:
Looks good. We'll update the `FEATURE` compose file when we're ready to migrate that to ECS.

PR #1708, opened 2017-03-23
added explicit separate compose file so ecs build can be done alongside prevous

PeterRJones:
#### Overview
To make cut-over to ecs-deployment adding both ecs-based and non-ecs-based docker-compose file. Main difference is the usage of dynamic host ports in the ecs case. 

deployment

#### Breaking Changes
None

#### Requirements
None

#### Other information
future PR will remove non-ecs files when cut-over is complete

### DIFF
--- a/docker-compose-quickstart-ecs.yml
+++ b/docker-compose-quickstart-ecs.yml
@@ -21,6 +21,7 @@ webapp:
   command:
     npm run server-auth
   log_driver: json-file
-  restart: always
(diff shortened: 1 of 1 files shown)


PeterRJones:
//cc @dhylbert

dhylbert:
I don't think we need this. If we need to fall back to the non-ECS deployment, we should override `CFN_PARAM_COMPOSE_FILE` in the `dev` jenkins job.

```
CFN_PARAM_COMPOSE_FILE="docker-compose-quickstart.yml -f docker-compose.override.yml"
```

PR #1750, opened 2017-04-05
Fix Commons CSS + serving of webpacked files

maxbates:
#### Overview

- fix commons build path, so CSS is served properly
- update server static routing
  - env var not set for server, so update serving of static bundles (built with webpack)
  - minor tweaks to webpack bundling file location
- Open commons via context menu
- commons header now says "sign in" instead of user glyph

fix

#### Breaking Changes

na

#### Requirements

- [x] Docs updated (for features + fixes)


maxbates:
Included a bundled file, that is now ignored. see #1751

PR #1775, opened 2017-04-13
Blast Menu UI Tweaks

justinmc:
Issue: https://github.com/Autodesk/genetic-constructor/issues/1758

This PR makes a few tweaks to the plugins context menu to make it match the spec in [zeplin](https://app.zeplin.io/project.html#pid=581d0bc0b7502238616341cb&sid=58a4ea278d3c3a5d387d7986).

screen shot 2017-04-13 at 2 22 35 pm

### DIFF
--- a/src/components/Menu/MenuItem.js
+++ b/src/components/Menu/MenuItem.js
@@ -34,11 +34,16 @@ export default class MenuItem extends Component {
     action: PropTypes.func,
     disabled: PropTypes.bool,
     checked: PropTypes.bool,
+    indent: PropTypes.bool,
     shortcut: PropTypes.string,
     classes: PropTypes.string,
     close: PropTypes.func.isRequired,
     menuItems: PropTypes.array,
     openLeft: PropTypes.bool,
+  };
+
+  static defaultProps = {
+    indent: false,
   };
 
   constructor() {
@@ -74,7 +79,7 @@ export default class MenuItem extends Component {
 
   render() {
     // indent if check able regardless of checked state
-    const indent = this.props.checked === true || this.props.checked === false;
+    const indent = this.props.indent || this.props.checked === true || this.props.checked === false;
     let check = null;
     if (indent) {
       check = <div className={this.props.checked ? 'menu-item-checked' : 'menu-item-unchecked'} />;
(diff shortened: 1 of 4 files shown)


justinmc:
Sorry, this isn't quite ready for review yet.  I didn't realize mentionbot was going to jump in and tag people :)

src/components/Menu/MenuItem.js
```diff
@@ -34,13 +34,18 @@ export default class MenuItem extends Component {
     action: PropTypes.func,
     disabled: PropTypes.bool,
     checked: PropTypes.bool,
+    indent: PropTypes.bool,
```
justinmc: Are we ok with adding an indent prop here?  Any better way to do this?

src/graphics/views/constructviewer.js
```diff
@@ -417,11 +417,15 @@ export class ConstructViewer extends Component {
 
     const extensionsWithBlockMenus = extensionsByRegion('menu:block')
     .filter(manifest => manifest.render['menu:block'])
-    .map(manifest => manifest.render['menu:block'](singleBlock, firstBlock));
+    .map(manifest =>
+      manifest.render['menu:block'](singleBlock, firstBlock)
+      // Indent all extension menuItems
+      .map(menuItem => Object.assign({}, menuItem, { indent: true })),
```
justinmc: I add `intent: true` to all extension menuItems, which I thought was more robust than just adding it to [extensions/blast/blockMenuItem.js](https://github.com/Autodesk/genetic-constructor/blob/master/extensions/blast/blockMenuItem.js).  Let me know if that's not true or if there's a better place to do this.

justinmc: I really wanted to add a quick spec for this just to check that all extension menu items have `indent: true`, but it looks like that's going to be a big undertaking.  This component doesn't have a spec file yet, and I would have to write a fake store factory or something like that to get it to render in a test.

I could split this `blockContextMenuItems` method out into a utils file or something if we want, that would be much more testable.

maxbates: none of the components have tests right now.... they are just implicitly tested by the end to end tests

